PR #1, opened 2013-12-19
do not set alloc number if 1

forderud:
Please (cherry-pick) merge this change to the "master" branch, so that it becomes part of the next boost release.

As of today, boost test always overrides the allocation number to break on. This is not a good idea if the user has already set a value from the debugger.

REF: http://lists.boost.org/Archives/boost/2013/09/206204.php


raffienficiaud:
Coming back to this: it is already in boost 1.59.

PR #9, opened 2014-06-17
[winrt support] Disabling usage of some banned APIs for the Windows Runtime

stgates:
Adding a macro for turning on/off for debugger support. Attaching a debugger is disabled if running tests for the Windows Runtime.

Disabled some usage of environment variables, which aren't avaliable, when under the Windows Runtime.

Replaced one usage of CreateFileA, which isn't allow in the Windows Runtime with fopen.

### DIFF
--- a/include/boost/test/debug.hpp
+++ b/include/boost/test/debug.hpp
@@ -1,5 +1,4 @@
 //  (C) Copyright Gennadiy Rozental 2006-2012.
-//  (C) Copyright Microsoft Corporation 2014.
 //  Distributed under the Boost Software License, Version 1.0.
 //  (See accompanying file LICENSE_1_0.txt or copy at 
 //  http://www.boost.org/LICENSE_1_0.txt)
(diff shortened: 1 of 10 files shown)


jneidlinger:
Pinging! Any chance this work could be revived? I would LOVE to be using Boost UTF on my Windows Runtime projects.

stgates:
@jneidlinger These changes here were to enable running some of the Boost tests covering the Windows Runtime. I'm not entirely sure how easy it would be to use for testing arbitrary Windows Runtime projects. 

More of the Windows APIs are being unbanned and allowed in the Windows Runtime, for example the socket and threading APIs. I'm canceling this pull request, as I think the changes are a bit invasive and going to be difficult to maintain.

jneidlinger:
@stgates I was more interested in using the same Boost tests we currently use to test some cross platform native libraries we share between Windows apps and WinRT apps.

Are you thinking enough banned APIs will be opened up to use Boost UTF out of the box (Windows) in the future?

Thanks for the update, I appreciate it.

stgates:
@jneidlinger I don't know if that will happened, but I don't have any more time to spend on this pull request. Also without regular testing it would be quite easy to break this work.

PR #15, opened 2014-10-17
Fixes for #7397 - Boost.Test, since boost 1.48 is using the deprecated Boost.Timer class - it should be updated to use the new class

ja11sop:
Pull request for https://svn.boost.org/trac/boost/ticket/7397 allowing boost.test to be used boost.timer. Please see the referenced ticket #7397 for full details.


ja11sop:
Apologies for the delay in this - I've been a little snowed under recently but will hopefully get back to this soon

raffienficiaud:
Sure, we also were quite busy lately

ja11sop:
Hi Guys, please review the updated pull request. Here is a summary:

1. The macro `BOOST_TEST_USE_DEPRECATED_TIMER` has been replaced by a runtime cla `--deprecated_timer_format`. Specifying this will give the old HRF output for the deprecated Boost.Timer

2. The XML output has been updated. Given the following test times 0.527163615s wall, 0.450000000s user, 0.050000000s system (and cpu time = user+system = 0.500000000s) the original patch would output the following times - *all in nanoseconds*. (My earlier comment regarding the XML output was incorrect - regardless though the suggested change still applies)
```
<TestingTime>500000000</TestingTime>
<WallTime>527163615</WallTime>
<UserTime>450000000</UserTime>
<SystemTime>050000000</SystemTime>
```
The updated patch will now output:
```
<TestingTime>500000</TestingTime>
<CpuTime>500000000</CpuTime>
<WallTime>527163615</WallTime>
<UserTime>450000000</UserTime>
<SystemTime>50000000</SystemTime>
```
That is, `<TestingTime>` will remain as it is today, a count in *microseconds* (the HRF output checks if the value is %1000 and displays ms instead of microseconds whereas the XML output just outputs the raw microsecond count). The new fields will all display their values in *nanoseconds* per the native output from the new Boost.Timer.

This means that existing parsers or scripts that handle the XML output as before (with the exception that Windows users get a more useful time). HRF users relying on the old format need to either update their scripts or enable the option.

raffienficiaud:
Hi Jamie,

Thanks for your continued efforts. I am happy with the features you implemented. I will review and test it tomorrow. We noticed we changed a lot of files in between, hopefully the merge went well.

I'll get back to you tomorrow (France is in mourning today).

Raffi

ja11sop:
No problem, take your time and thanks for the quick reply.

raffienficiaud:
Hi,

I merged your pull request to the branch 

I have the following errors now when I try to build and run the unit test. I think it is in connection with the inclusion of `boost/chrono`:

```
darwin.compile.c++ ../../../bin.v2/libs/test/test/basic_cstring_test.test/darwin-4.2.1/debug/basic_cstring_test.o
In file included from basic_cstring_test.cpp:21:
In file included from ../../../boost/test/unit_test.hpp:18:
In file included from ../../../boost/test/test_tools.hpp:31:
In file included from ../../../boost/test/tools/old/impl.hpp:19:
In file included from ../../../boost/test/unit_test_log.hpp:18:
In file included from ../../../boost/test/tree/observer.hpp:20:
In file included from ../../../boost/test/utils/timer.hpp:15:
In file included from ../../../boost/timer/timer.hpp:14:
In file included from ../../../boost/chrono/chrono.hpp:13:
In file included from ../../../boost/chrono/system_clocks.hpp:64:
In file included from ../../../boost/chrono/detail/system.hpp:12:
In file included from ../../../boost/system/error_code.hpp:14:
../../../boost/system/config.hpp:34:3: error: Must not define both BOOST_SYSTEM_DYN_LINK and BOOST_SYSTEM_STATIC_LINK
# error Must not define both BOOST_SYSTEM_DYN_LINK and BOOST_SYSTEM_STATIC_LINK
  ^
1 error generated.

    "g++"  -ftemplate-depth-128 -O0 -fno-inline -Wall -pedantic -g -dynamic -gdwarf-2 -fexceptions -Wno-long-long -fPIC  -DBOOST_ALL_NO_LIB=1 -DBOOST_CHRONO_DYN_LINK=1 -DBOOST_CHRONO_STATIC_LINK=1 -DBOOST_CHRONO_THREAD_DISABLED -DBOOST_SYSTEM_DYN_LINK=1 -DBOOST_SYSTEM_NO_DEPRECATED -DBOOST_SYSTEM_STATIC_LINK=1 -DBOOST_TEST_NO_AUTO_LINK=1 -DBOOST_TIMER_DYN_LINK=1 -DBOOST_TIMER_STATIC_LINK=1  -I"../../.." -c -o "../../../bin.v2/libs/test/test/basic_cstring_test.test/darwin-4.2.1/debug/basic_cstring_test.o" "basic_cstring_test.cpp"
```

Any idea?
Thanks,
Raffi

ja11sop:
Yes - darwin is a little tricky - basically it looks like Boost.Test either adds the static link defines, or adds the dynamic link all macro. That shouldn't happen. Since previously boost.test did not rely on Boost.Timer and hence Boost.System is might have an override that just worked? I don't have darwin here to try this. What's your b2 invocation line? I might be able to try this here on linux.

raffienficiaud:
Maybe it is reproducible on Linux, because I think in the jamfile the static lib is explicitly required sometimes. I just run b2 from the `boost/libs/test/test` folder. The command should build and run the necessary.

ja11sop:
Ah ok, so nothing special with b2 then. I'll give it a go and get back to you. If you prefer we can take this off the ticket and I can email you directly. Either way though I'll post back here after I've tried to reproduce.

ja11sop:
Sorry for the slow reply - I got side-tracked. I can replicate this so I'lI investigate.

raffienficiaud:
Would the pull request #29 solve the issue?

ja11sop:
So I've just arrived here to make the very suggestion in  #29 - It was exactly what I was about to suggest but I see I have been beaten to it, so yes please go ahead.

ja11sop:
Jürgen's pull request is now applied - I somehow missed that earlier.

raffienficiaud:
I pulled it into the same branch `tractickets/7397`. Now having some other issues on OSX (apart from some UT that are now failing because of the change of format, which I will fix later). The header variant seems broken. 

```
darwin.compile.c++ ../../../bin.v2/libs/test/test/single_header_test.test/darwin-4.2.1/debug/single_header_test.o
darwin.link ../../../bin.v2/libs/test/test/single_header_test.test/darwin-4.2.1/debug/single_header_test
Undefined symbols for architecture x86_64:
  "boost::timer::format(boost::timer::cpu_times const&, short, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)", referenced from:
      boost::unit_test::to_string(boost::timer::cpu_times) in single_header_test.o
  "boost::timer::cpu_timer::stop()", referenced from:
      boost::unit_test::framework_impl::visit(boost::unit_test::test_case const&) in single_header_test.o
  "boost::timer::cpu_timer::start()", referenced from:
      boost::timer::cpu_timer::cpu_timer() in single_header_test.o
  "boost::system::system_category()", referenced from:
      ___cxx_global_var_init2 in single_header_test.o
  "boost::system::generic_category()", referenced from:
      ___cxx_global_var_init in single_header_test.o
      ___cxx_global_var_init1 in single_header_test.o
  "boost::timer::cpu_timer::elapsed() const", referenced from:
      boost::unit_test::framework_impl::visit(boost::unit_test::test_case const&) in single_header_test.o
ld: symbol(s) not found for architecture x86_64
```

rogeeff:
Should we CC Jurgen?

On Sat, Jan 10, 2015 at 3:21 PM, Raffi Enficiaud <notifications@github.com>
wrote:

> I pulled it into the same branch tractickets/7397. Now having some other
> issues on OSX (apart from some UT that are now failing because of the
> change of format, which I will fix later). The header variant seems broken.
>
> darwin.compile.c++ ../../../bin.v2/libs/test/test/single_header_test.test/darwin-4.2.1/debug/single_header_test.o
> darwin.link ../../../bin.v2/libs/test/test/single_header_test.test/darwin-4.2.1/debug/single_header_test
> Undefined symbols for architecture x86_64:
>   "boost::timer::format(boost::timer::cpu_times const&, short, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)", referenced from:
>       boost::unit_test::to_string(boost::timer::cpu_times) in single_header_test.o
>   "boost::timer::cpu_timer::stop()", referenced from:
>       boost::unit_test::framework_impl::visit(boost::unit_test::test_case const&) in single_header_test.o
>   "boost::timer::cpu_timer::start()", referenced from:
>       boost::timer::cpu_timer::cpu_timer() in single_header_test.o
>   "boost::system::system_category()", referenced from:
>       ___cxx_global_var_init2 in single_header_test.o
>   "boost::system::generic_category()", referenced from:
>       ___cxx_global_var_init in single_header_test.o
>       ___cxx_global_var_init1 in single_header_test.o
>   "boost::timer::cpu_timer::elapsed() const", referenced from:
>       boost::unit_test::framework_impl::visit(boost::unit_test::test_case const&) in single_header_test.o
> ld: symbol(s) not found for architecture x86_64
>
> —
> <https://github.com/boostorg/test/pull/15#issuecomment-69470420>.
>



-- 
Gennadiy Rozental

raffienficiaud:
He is already as participant of the PR.

jhunold:
And watching the repository :-) Will take a look tomorrow.

raffienficiaud:
I feel like someone breathing over my shoulder :)

ja11sop:
Hi Raffi, try out my latest change to add Boost.Timer as a library needed by the `single_header_test`. It resolves the issue you are seeing for me using g++. Jamie

raffienficiaud:
I solves the problem, but I think the problem lies somewhere else. Anyway, for your fix, I need to solve the UT that is failing now and it should be good.

ja11sop:
Yes - I think I know the issue with the failing UT - will let you know once I have a working patch

ja11sop:
The issue with the failing unit test was that the old timer (having a lower accuracy) would typically return a 0 time for tests that are executed immediately. In the case of the HRF output used in the pattern matching test it relies on there being no `; testing time ...` appended to the `Leaving test <test_type> ...` message.

Since the pattern matching isn't actually pattern matching at all but a simple by-character comparison it's not feasible to somehow match against a time pattern. It seemed that the best approach was to allow suppression of the timer output altogether in the case of the failing test. In fact I think that's a generally useful feature for people you want to write simple comparison scripts against the test output.

I therefore added a new runtime parameter (and docs for it) called `--suppress_timer_output`. If this is specified no test time output is provided. I also fixed a copy-paste error for `--deprecated_timer_format`. 

What I haven't done, but we could consider, is update the test helper in the Jamfile to allow test arguments to be specified. Currently I just add `--suppress_timer_output` directly. So instead of this:
```
test-btl-lib ( test-rule : test-name : lib-name ? : pattern_file * : source_files * : extra-libs ? : extra-options ? )
```
we might instead have this
```
test-btl-lib ( test-rule : test-name : args* :  lib-name ? : pattern_file * : source_files * : extra-libs ? : extra-options ? )
```
and then we would only specify the option for the failing test.

raffienficiaud:
Hi,
I do not think this option is really useful: having accurate timing in logs and yet doing char to char comparison is just a non-sense and a limitation of the output comparison. I would go for running the tests with the old format instead. 
What do you think?

ja11sop:
Well - I agree, that would have been my preference. The issue is that a non-zero time may be displayed, regardless of format. The test relies on there being no time displayed. Having said that I'll run it a few times here and see if it will do what we want regardless.

ja11sop:
Quick update - using the old format is not sufficient - the issue is we still produce a displayed time. It should probably be noted that suppressing the timer output is only to allow the test output comparison to be checked.

I wouldn't consider it to be an option you'd not generally, if ever, want to use, except of course in the case of classic test output comparisons. Then you don't really want a timer causing an unwanted diff. In that sense I think the option is not unreasonable, but I do agree it would be nicer if the output comparison test itself was not quite so brittle,

I suspect I ran into some of these issues when I first put the patch together in the SVN days and it probably influenced my macro based approach. Boost.Test though has moved on quite a lot since then so it is reasonable there will be some issues like this now.

Thinking on this more and looking at the code and intent of `match_pattern()` in [test_tools.ipp](https://github.com/boostorg/test/blob/tractickets/7397/include/boost/test/impl/test_tools.ipp#L574) I think suppressing the time output is possibly the correct approach. That we had no time output before was by luck and not by design. Further it appears that the intent of the matching function is to compare two files for an exact match. 'pattern' being a misnomer in this case.

Either way, I think this needs more thought, and it might be that despite an initial dislike of the approach it may in fact be the best approach.

ja11sop:
Raffi, it turns out I can make this work with only --deprecated_timer_format by updating `has_time()` to only consider the User and System time in that case. Unless you've decided you like the utility of `--suppress_timer_output` I'll go ahead and update the patch to remove the additional option. Also I think it will make sense to also update the `test-btl-lib` rules to take arguments to make the Jamfile cleaner. Let me know your thoughts. Jamie

ja11sop:
I've gone ahead and removed the `--suppress_timer_output` option. If we decide we want something like that later it will be easy to re-add it. Give the current pull-request a run. All tests pass as expected for me (now I have a proper modular boost setup) - Jamie

raffienficiaud:
Good, I'll test that. I was wondering why with the old format the unit tests were failing, but you gave me the answer.

raffienficiaud:
Works.

ja11sop:
Excellent!

raffienficiaud:
And merged, thanks again for working on this.

ja11sop:
No problem - thanks for taking the time to give it careful consideration. I'd very much appreciate you also looking at my other two pull requests when you have the time. I've double-checked #16 and it is good, but I still need to review and update #17, especially in light of the new `--deprecated_timer_format` option. I'll update that pull request when I've done so.

rogeeff:
I've commented on #16

On Mon, Jan 12, 2015 at 9:03 AM, ja11sop <notifications@github.com> wrote:

> No problem - thanks for taking the time to give it careful consideration.
> I'd very much appreciate you also looking at my other two pull requests
> when you have the time. I've double-checked #16
> <https://github.com/boostorg/test/pull/16> and it is good, but I still
> need to review and update #17 <https://github.com/boostorg/test/pull/17>,
> especially in light of the new --deprecated_timer_format option. I'll
> update that pull request when I've done so.
>
> —
> <https://github.com/boostorg/test/pull/15#issuecomment-69573530>.
>



-- 
Gennadiy Rozental

raffienficiaud:
This need to be reworked as boost.timer is creating a lot of issues in other libraries.

ja11sop:
Ok, I can imagine that might be the case given the ultimate dependency on system and chrono. If you want to point me in the right direction of some of the issues I'll take a look and work with you to find a solution. What's your general feeling about this as a whole?

rogeeff:
General feeling is that we should have tested this better and develop
branch is not a good place for experiments. Plus I have number of style
issues with these changes I am about to push.

As for the fix, we should support purely header only option. What is the
easiest route for this is the one we should take.

On Wed, Jan 14, 2015 at 4:33 PM, ja11sop <notifications@github.com> wrote:

> Ok, I can imagine that might be the case given the ultimate dependency on
> system and chrono. If you want to point me in the right direction of some
> of the issues I'll take a look and work with you to find a solution. What's
> your general feeling about this as a whole?
>
> —
> <https://github.com/boostorg/test/pull/15#issuecomment-69996252>.
>



-- 
Gennadiy Rozental

rogeeff:
I mean, I am about to push fixes

On Wed, Jan 14, 2015 at 5:39 PM, Gennadiy Rozental <rogeeff@gmail.com>
wrote:

> General feeling is that we should have tested this better and develop
> branch is not a good place for experiments. Plus I have number of style
> issues with these changes I am about to push.
>
> As for the fix, we should support purely header only option. What is the
> easiest route for this is the one we should take.
>
> On Wed, Jan 14, 2015 at 4:33 PM, ja11sop <notifications@github.com> wrote:
>
>> Ok, I can imagine that might be the case given the ultimate dependency on
>> system and chrono. If you want to point me in the right direction of some
>> of the issues I'll take a look and work with you to find a solution. What's
>> your general feeling about this as a whole?
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/boostorg/test/pull/15#issuecomment-69996252>.
>>
>
>
>
> --
> Gennadiy Rozental
>



-- 
Gennadiy Rozental

raffienficiaud:
Fixes of? I just reverted the merge.

rogeeff:
hmm.. ok. We'll have to review this more carefully before we merge it again.

On Wed, Jan 14, 2015 at 5:46 PM, Raffi Enficiaud <notifications@github.com>
wrote:

> Fixes of? I just reverted the merge.
>
> —
> <https://github.com/boostorg/test/pull/15#issuecomment-70008037>.
>



-- 
Gennadiy Rozental

ja11sop:
Apart from stylistic issues I assume the merge is causing failures in other libraries? Or is this primarily a build issue (libraries depend on test, but test now depends on timer, system and chrono) which is certainly a  problem (though I've not seen it in *using* boost for the past few years). My previous fallback was of course to allow using the original boost.timer. If the build dependencies (and issues) are insurmountable it might be feasible to take a step back and use a small, lightweight header-only equivalent to boost.timer - removing the dependency entirely.

raffienficiaud:
Yes, in almost everything that is using boost.test as header only, or is using boost.timer old API. I should have taken a deeper look to the boost.timer new vs. old API anyway. I am also reluctant in the fallback macro as we have to maintain two branch of compilation and we do not get rid of the initial problem, which was:
- having a more accurate timing in boost.test
- getting rid of a deprecated API

All the changes remain in the branch tracticket/7397, and I propose we put these things on hold for a moment.

ja11sop:
Agreed. I'll take a look at a lightweight header only option, but yes, putting this on hold for the moment is a good idea.

bchretien:
Any update on this? And in the meantime, is there a workaround for this issue?

raffienficiaud:
Unfortunately no. I involves some changes in the boost.timer library to support header only inclusion.

PR #18, opened 2014-10-23
Tests & fixes for Boost.Test UTF command-line parsing.

jurko-gospodnetic:
UTF seems to have some serious issues with its command-line parsing. :-(

This pull request adds a test module for it, and fixes one of the problems - a segfault that occurs if you attempt to pass it an argument with a leading embedded space character or an empty string as the initial but not only argument.

There are other problems with this command-line handling, but I'll add failing tests for some of them as a separate pull request based on this one.

Best regards,
  Jurko Gospodnetić


rogeeff:
I am completely rewriting CLA support for next release. It'll address lots of issues including these.

PR #29, opened 2015-01-10
Fix Boost.Build issues

jhunold:
Relevant commit is e2fd5e8

I've removed the explicit linking to Boost.System and Boost.Timer as both dependencies are handled by Boost.Build automagically.  The test basic_cstring_test
         [ test-btl-lib run : basic_cstring_test    : boost_unit_test_framework/<link>static ]
 requires static runtime. Unconditionally adding both libraries globally (for all tests) results in linking against both shared and static library.
And you only need <use> instead of <library> to draw header dependencies.


raffienficiaud:
Sorry, I do not get it. Why are you closing the pull request?

jhunold:
I wanted a clean patch against the original pull request. See https://github.com/ja11sop/test/pull/1 for this. Still experimenting with GitHub.

raffienficiaud:
Since I did not understand exactly what happened, I merged the result of the merge of ja11sop#1 instead into the branch `tractickets/7397`.

I now have a problem with the header only variant of the unit test. However we can close this PR.

Thanks,
Raffi

PR #44, opened 2015-03-29
Fix compilation errors on msvc caused by min/max macros.

awulkiew:


raffienficiaud:
Thanks!
Would you please point me to the failing tests?

awulkiew:
See e.g. distance_x_x Geometry tests for msvc in the develop regression matrix (http://www.boost.org/development/tests/develop/developer/geometry.html).

E.g. http://www.boost.org/development/tests/develop/developer/output/teeks99-08e-win2012R2-64on64-boost-bin-v2-libs-geometry-test-algorithms-distance-distance_areal_areal-test-msvc-12-0-dbg-adrs-mdl-64-archt-x86-async-excpt-on-thrd-mlt.html

raffienficiaud:
Thanks, I will add a test to catch this.

PR #52, opened 2015-05-05
Fix feature test of BOOST_NO_CXX11_VARIADIC_MACROS

jessicah:
Change b41a9354e9127e77d19a1b5c975c33729b033218 introduced regression failures due to preprocessor errors with respect to the && operator, leading to the following error:

../boost/test/test_tools.hpp:50:87: error: operator '&&' has no right operand
 #if !BOOST_PP_VARIADICS || ((__cplusplus >= 201103L) && BOOST_NO_CXX11_VARIADIC_MACROS)


raffienficiaud:
Applied, thanks!

PR #58, opened 2015-05-19
Add needed MPL header file.

eldiener:
The MPL header file needs to be correctly included. This is necessary for upcoming type_traits version 2, which does not depend on MPL anymore.


raffienficiaud:
Thanks!

PR #64, opened 2015-05-22
Doc: more on test case ordering

akrzemi1:


doc/test_organization/managing_tests_dependencies.qbk
```diff
@@ -7,17 +7,31 @@
 
 [section:tests_dependencies Managing test dependencies]
 
-Dependencies among tests are managed by means of [link boost_test.tests_organization.decorators decorators]. 
+In general, you should make sure that test cases are independent on one another.
```
raffienficiaud: why?

doc/test_organization/managing_tests_dependencies.qbk
```diff
@@ -7,17 +7,31 @@
...
+For any two test cases `TA` and `TB`, `TB` must not take for granted that `TA` has already executed,
+even if `TA` is declared before `TB` in the same translation unit.
+
+The only ordering-related guarantee that __UTF__ makes by default is that if test cases `TA` and `TB`
```
raffienficiaud: I would just say that there is no order guarantied bu the framework/test runner, except otherwise specified by the dependency declaration

doc/test_organization/managing_tests_dependencies.qbk
```diff
@@ -7,17 +7,31 @@
...
+are declared in the same test suite, no test case (call it `TX`) from any other test suite is executed between `TA` and `TB`,
+even if the declaration of `TX` appears between the declarations of `TA` and `TB`.
+In other words, all tests from a suite are executed in one go, even if the test suite namespace is opened multiple times.
```
raffienficiaud: Is this really a guaranty? personnally I do not know (so I'm not questioning your sentence but the framework)

rogeeff: Yes this is true

doc/test_organization/managing_tests_dependencies.qbk
rogeeff: Because test case independence is corner stone of TDD

On Fri, May 22, 2015 at 5:51 PM, Raffi Enficiaud <notifications@github.com>
wrote:

> In doc/test_organization/managing_tests_dependencies.qbk
> <https://github.com/boostorg/test/pull/64#discussion_r30933423>:
>
> > @@ -7,17 +7,31 @@
> >
> >  [section:tests_dependencies Managing test dependencies]
> >
> > -Dependencies among tests are managed by means of [link boost_test.tests_organization.decorators decorators].
> > +In general, you should make sure that test cases are independent on one another.
>
> why?
>
> —
> <https://github.com/boostorg/test/pull/64/files#r30933423>.
>



-- 
Gennadiy Rozental

raffienficiaud: I would rather put the emphasis on the order: users should not rely on any execution order, and tests execution can be shuffled.

doc/test_organization/managing_tests_dependencies.qbk
akrzemi1: 2015-05-22 23:53 GMT+02:00 Raffi Enficiaud <notifications@github.com>:

> In doc/test_organization/managing_tests_dependencies.qbk
> <https://github.com/boostorg/test/pull/64#discussion_r30933561>:
>
> > @@ -7,17 +7,31 @@
> >
> >  [section:tests_dependencies Managing test dependencies]
> >
> > -Dependencies among tests are managed by means of [link boost_test.tests_organization.decorators decorators].
> > +In general, you should make sure that test cases are independent on one another.
> > +For any two test cases `TA` and `TB`, `TB` must not take for granted that `TA` has already executed,
> > +even if `TA` is declared before `TB` in the same translation unit.
> > +
> > +The only ordering-related guarantee that __UTF__ makes by default is that if test cases `TA` and `TB`
> > +are declared in the same test suite, no test case (call it `TX`) from any other test suite is executed between `TA` and `TB`,
> > +even if the declaration of `TX` appears between the declarations of `TA` and `TB`.
> > +In other words, all tests from a suite are executed in one go, even if the test suite namespace is opened multiple times.
>
> Is this really a guaranty? personnally I do not know (so I'm not
> questioning your sentence but the framework)
>

I think we need to guarantee that in order for the "suite entry/exit
fixture" to make sense. This is what the entry/exit fixture guarantees:

first suite setup, then all suite tests, last suite tear down. So, this
precludes any shuffling of tests between suites (but allows shufflig within
one suite).

doc/test_organization/managing_tests_dependencies.qbk
akrzemi1: 2015-05-23 10:16 GMT+02:00 Raffi Enficiaud <notifications@github.com>:

> In doc/test_organization/managing_tests_dependencies.qbk
> <https://github.com/boostorg/test/pull/64#discussion_r30943082>:
>
> > @@ -7,17 +7,31 @@
> >
> >  [section:tests_dependencies Managing test dependencies]
> >
> > -Dependencies among tests are managed by means of [link boost_test.tests_organization.decorators decorators].
> > +In general, you should make sure that test cases are independent on one another.
>
> I would rather put the emphasis on the order: users should not rely on any
> execution order, and tests execution can be shuffled.
>

Are you suggesting to only change the wording from imperative "you should
not make the tests dependent" to a more informative "you should not rely on
the ordering"?

If so, I do not see any problem with such change.

Regards,
&rzej

raffienficiaud: Yes. The link between test dependency and execution order is also not obvious.

akrzemi1: 2015-05-25 10:20 GMT+02:00 Raffi Enficiaud <notifications@github.com>:

> In doc/test_organization/managing_tests_dependencies.qbk
> <https://github.com/boostorg/test/pull/64#discussion_r30968211>:
>
> > @@ -7,17 +7,31 @@
> >
> >  [section:tests_dependencies Managing test dependencies]
> >
> > -Dependencies among tests are managed by means of [link boost_test.tests_organization.decorators decorators].
> > +In general, you should make sure that test cases are independent on one another.
>
> Yes. The link between test dependency and execution order is also not
> obvious.
>
I do not understand what you are saying with this. Being ordered before
some other test is being dependent on it, isn't it?

doc/test_organization/managing_tests_dependencies.qbk
raffienficiaud: To me, these are orthogonal:
- you make a test entry/exit to ensure that every test under the test suite have executed this fixture once (which is different from have executed this fixture "always")
- the fact that we execute all test units under a test suite in one go can be considered as a minimal number of execution for this suite entry/exit fixtures, eg if the fixture one is costly. This is somehow not the case when shuffling (I guess).

doc/test_organization/managing_tests_dependencies.qbk
raffienficiaud: They can be ordered without dependency, aren't they? Dependency enforce order, the opposite is not true.

akrzemi1: 2015-05-25 10:32 GMT+02:00 Raffi Enficiaud <notifications@github.com>:

> In doc/test_organization/managing_tests_dependencies.qbk
> <https://github.com/boostorg/test/pull/64#discussion_r30968769>:
>
> > @@ -7,17 +7,31 @@
> >
> >  [section:tests_dependencies Managing test dependencies]
> >
> > -Dependencies among tests are managed by means of [link boost_test.tests_organization.decorators decorators].
> > +In general, you should make sure that test cases are independent on one another.
>
> They can be ordered without dependency, aren't they? Dependency enforce
> order, the opposite is not true.
>
Ok, I think I see your point. You want to use term "dependency" to talk
about the union of "ordered before" and "run only if dependency
successful". This means that the section on ordering only should go
elsewhere, right?

rogeeff: I did not have a chance to read through this. Here few points I'd like to
make:


   1. Default order follows order of test unit declaration in translation
   unit
   2. There is no promise bet ween translation units
   3. All test units within the same test suite are executed together
   4. Dependency enforces order, not the other way around
   5. random test unit execution shuffles test siblings of *all* test
   suites independently.
   6. Most complicated to explain to the case of dependency + random. What
   happens is this: while analyzing the dependency we assign "rank" to all
   siblings among test suite children. Lets say we have 2 children with rank 1
   - tu1 and tu2 and 2  children with rank 2 - tu3 and tu4. We will ran first
   tu1 and tu2 in random order, followed by tu3 and tu4 in random order


On Mon, May 25, 2015 at 4:32 AM, Raffi Enficiaud <notifications@github.com>
wrote:

> In doc/test_organization/managing_tests_dependencies.qbk
> <https://github.com/boostorg/test/pull/64#discussion_r30968769>:
>
> > @@ -7,17 +7,31 @@
> >
> >  [section:tests_dependencies Managing test dependencies]
> >
> > -Dependencies among tests are managed by means of [link boost_test.tests_organization.decorators decorators].
> > +In general, you should make sure that test cases are independent on one another.
>
> They can be ordered without dependency, aren't they? Dependency enforce
> order, the opposite is not true.
>
> —
> <https://github.com/boostorg/test/pull/64/files#r30968769>.
>



-- 
Gennadiy Rozental

raffienficiaud: My view is that all of those orders are implementation details. If some test should be run after another one, then the user should use depends_on decorator. If the suite level entry/exit fixture should be detailed (in terms of number of execution of the fixture itself), I think this is enough to say that the test tree is processed in a breadth first manner, which is exactly that. When I shuffle the test, I am expecting that there is no order at all, and test cases are picked randomly. If a test case is to be executed, and if I have random, I am (as a user) expecting that the execution is postponed until all parent test cases are executed (which is not the same as ordering by rank). So the algorithm should look like this:

    while not queue.empty():
          current_test_case = queue.pop()
          if not current_test_case.all_parents_executed():
              queue.push_back(current_test_case)
              continue
          
          # execute test 

Here, with random, I am expecting that queue has been shuffled.

akrzemi1: 2015-05-26 10:09 GMT+02:00 Raffi Enficiaud <notifications@github.com>:

> In doc/test_organization/managing_tests_dependencies.qbk
> <https://github.com/boostorg/test/pull/64#discussion_r31011991>:
>
> > @@ -7,17 +7,31 @@
> >
> >  [section:tests_dependencies Managing test dependencies]
> >
> > -Dependencies among tests are managed by means of [link boost_test.tests_organization.decorators decorators].
> > +In general, you should make sure that test cases are independent on one another.
>
> My view is that all of those orders are implementation details. If some
> test should be run after another one, then the user should use depends_on
> decorator. If the suite level entry/exit fixture should be detailed (in
> terms of number of execution of the fixture itself), I think this is enough
> to say that the test tree is processed in a breadth first manner, which is
> exactly that. When I shuffle the test, I am expecting that there is no
> order at all, and test cases are picked randomly. If a test case is to be
> executed, and if I have random, I am (as a user) expecting that the
> execution is postponed until all parent test cases are executed (which is
> not the same as ordering by rank). So the algorithm should look like this:
>
> while not queue.empty():
>       current_test_case = queue.pop()
>       if not current_test_case.all_parents_executed():
>           queue.push_back(current_test_case)
>           continue
>
>       # execute test
>
> Here, with random, I am expecting that queue has been shuffled.
>
> —
> <https://github.com/boostorg/test/pull/64/files#r31011991>.
>

So, you are saying two things:
1. This is technical detail, not suitable for the docs
2. The shuffling algorithm should be different than what it is now.

I ignore (1) for now, and for (2) I will let Gennadiy speak for this, but
my opinion is that it is not clear why your algorithm should be preferred
to the current one. The  current one already guarantees that (unless
depend_on) is used no two tests TA and TB will be ordered in the
predictable way.

raffienficiaud: Mmm, I am not sure, especially when I read "we assign "rank" to all siblings among test suite children." So there is still an ordering following the tree structure, while it should not be. This is why I am also proposing leaving out the (1).

akrzemi1: 2015-05-26 10:49 GMT+02:00 Raffi Enficiaud <notifications@github.com>:

> In doc/test_organization/managing_tests_dependencies.qbk
> <https://github.com/boostorg/test/pull/64#discussion_r31014678>:
>
> > @@ -7,17 +7,31 @@
> >
> >  [section:tests_dependencies Managing test dependencies]
> >
> > -Dependencies among tests are managed by means of [link boost_test.tests_organization.decorators decorators].
> > +In general, you should make sure that test cases are independent on one another.
>
> Mmm, I am not sure, especially when I read "we assign "rank" to all
> siblings among test suite children." So there is still an ordering
> following the tree structure, while it should not be. This is why I am also
> proposing leaving out the (1).
>
While I have not read the implementation, I am pretty sure that ranking
only reflects the depends_on relation and nothing more. (I.e, when you have
no depends_on, all ranks are same.) When I test this, the tests are
executed in the order that I find random enough.

I need to describe the ordering somewhere in order to:
* make it clear what --random is for
* indicate that *depends_on affects the ordering

I can move it to another place.

PR #69, opened 2015-05-26
Don't use std::numeric_limits when is_specialized is false

akrzemi1:
Prior to this change, in floating-point computations, in the case
where we work for types where std::numeric_limits are not specialized,
funcitons min() and max() from std::numeric_limits<FPT> are nonetheless
instantiated and return FPT(0). While the result is never used in run-time,
this may still trigger a compile-time failure.

After this change, the paths for the cases where std::numeric_limits are not
specialized are handled by separate specializations rather than run-time if
statements.

### DIFF
--- a/include/boost/test/tools/floating_point_comparison.hpp
+++ b/include/boost/test/tools/floating_point_comparison.hpp
@@ -23,6 +23,7 @@
 #include <boost/mpl/bool.hpp>
 #include <boost/type_traits/is_floating_point.hpp>
 #include <boost/type_traits/is_array.hpp>
+#include <boost/type_traits/conditional.hpp>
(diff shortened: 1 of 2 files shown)


rogeeff:
Why don't we merge fpc_limits instead of having the came code in two
places. Can fpc_op use version from other file.

akrzemi1:
> Why don't we merge fpc_limits instead of having the came code in two places. Can fpc_op use version from other file.

I am confused. What should we merge `fpc_limits` with? I understand that the goal of `fpc_op.hpp` is to select the right comparison algorithm and to report failure in a UTF format. It does not use any information on limits, I cannot see any overlap in code between the two files.

rogeeff:
sorry. Misread the diff

PR #73, opened 2015-05-29
Doc: improved FP comparison description

akrzemi1:


raffienficiaud:
Please update the jamfile as well

akrzemi1:
I have.

doc/testing_tools/testing_floating_points.qbk
```diff
@@ -135,29 +140,42 @@ inequations [link equ2 (2)] and [link equ3 (3)] where all underflow, overflow co
 || abs(u - v)/abs(v) <= epsilon; // (5)
 ``
 
+The above however, will not work when `v` or `u` is too close to zero. In such cases the solution is to resort to [link equ1 (1)]. 
+
 [h3 Implementation]
```
raffienficiaud: I added this part there when I fixed the support for multiprecision, but now I think it should go to another section since we are under "theory".

doc/testing_tools/testing_floating_points.qbk
```diff
@@ -83,23 +83,29 @@ BOOST_AUTO_TEST_CASE(test, * utf::tolerance(0.02))
...
+
+
+The following is the most obvious way to compare two floating-point values `u` and `v` for being close at a given absolute tolerance `epsilon`:
 
 [#equ1]
 ``
    abs(u - v) <= epsilon; // (1)
 ``
 
+However, in many circumstances, this is not what we want. A 1% tolerable difference between two huge numbers and a 1% tolerable difference
```
raffienficiaud: I think it is misleading to say "1%" here, because a percentage does not have any unit. The epsilon in the equation does have the same unit as u or v

doc/testing_tools/testing_floating_points.qbk
```diff
@@ -83,23 +83,29 @@ BOOST_AUTO_TEST_CASE(test, * utf::tolerance(0.02))
...
+between two tiny numbers make two different absolute tolerances. See [link Squassabia] for examples.
 
-does not produce expected results in many circumstances - specifically for very small or very big values (see
-[link Squassabia] for examples). The __UTF__ implements floating-point comparison algorithm that is
-based on the more confident solution first presented in [link KnuthII Knuth]:
+We do not want to apply the same absolute tolerance for huge and tiny numbers. Instead, we would like to scale the `epsilon` with `u` and `v`.
```
raffienficiaud: Equ4&5 are doing exactly the opposite: they are removing the scale from u and v rather than scaling epsilon

doc/testing_tools/testing_floating_points.qbk
```diff
@@ -83,23 +83,29 @@ BOOST_AUTO_TEST_CASE(test, * utf::tolerance(0.02))
...
+The __UTF__ implements floating-point comparison algorithm that is based on the solution first presented in [link KnuthII Knuth]:
```
raffienficiaud: I do not know if the Knuth is the first one presenting this

doc/testing_tools/testing_floating_points.qbk
```diff
@@ -135,29 +140,42 @@ inequations [link equ2 (2)] and [link equ3 (3)] where all underflow, overflow co
 || abs(u - v)/abs(v) <= epsilon; // (5)
 ``
 
+The above however, will not work when `v` or `u` is too close to zero. In such cases the solution is to resort to [link equ1 (1)].
```
raffienficiaud: "too close to 0" works, the problem is when one of the operand is really 0 (the Equations are not defined).

doc/testing_tools/testing_floating_points.qbk
```diff
@@ -135,29 +140,42 @@ inequations [link equ2 (2)] and [link equ3 (3)] where all underflow, overflow co
...
+The above however, will not work when `v` or `u` is too close to zero. In such cases the solution is to resort to [link equ1 (1)]. 
+
 [h3 Implementation]
 
-Checks based on equations [link equ4 (4)] and [link equ5 (5)] is implemented by the binary predicate 
-[classref boost::math::fpc::close_at_tolerance `close_at_tolerance`]. The method for comparison (Equ [link equ4 (4)] or [link equ5 (5)]) as well as the
-epsilon are selected at construction. This predicate reports the tolerance to set to make the test successful.
+Binary predicate [classref boost::math::fpc::close_at_tolerance `close_at_tolerance`] implements checks [link equ4 (4)] and [link equ5 (5)].
+The tolerance and the seleciotn between [link equ4 (4)] and [link equ5 (5)] is determined during construction. Apart from the Boolean result,
+the predicate can also return the minimum tolerance required to check positively the closeness of `u` and `v`.
```
raffienficiaud: "the minimum tolerance required to check positively ": this is incorrect and this is not what "the tolerance to set to make the test successful" says:
- the tolerance to set depends on the equation currently under check
- it also handles the inequality, while you are saying "check positively the closeness of `u` and `v`"

doc/testing_tools/testing_floating_points.qbk
```diff
@@ -135,29 +140,42 @@ inequations [link equ2 (2)] and [link equ3 (3)] where all underflow, overflow co
...
+
+Assertion __BOOST_TEST__ (when comparing floating-point numbers) uses the following algorithm:
+
+* When either `u` or `v` is zero, evaluates [classref boost::math::fpc::small_with_tolerance `small_with_tolerance`] on the other.
+* When the specified tolerance is zero, preforms direct (native) comparison between `u` and `v`.
+* Otherwise, performs [classref boost::math::fpc::close_at_tolerance `close_at_tolerance`] using [link equ4 (4)].
+
+[note Therefore in order to check if a number is close to zero with tolerance, you need to type: 
+```
+BOOST_TEST(v == T(0), tt::tolerance(eps));
```
raffienficiaud: Are we required to type `T(0)` ? This is the question I asked you for this test `0 < fpt` in another PR

doc/testing_tools/testing_floating_points.qbk
```diff
@@ -135,29 +140,42 @@ inequations [link equ2 (2)] and [link equ3 (3)] where all underflow, overflow co
...
-to detect that the type is suitable for comparisons using the tolerance scheme explained above. This detection involves
-`std::numeric_limits`. If it turns out that this automatic detection does not work for a specific type, it is then
-possible to specialise the meta-class [classref boost::math::fpc::tolerance_based] for this specific type to derive from 
-`boost::mpl::true_`.
+* ['less-at-tolerance] <==> ['strictly-less] and not ['close-at-tolerance]
+* ['greater-at-tolerance] <==> ['strictly-greater] and not ['close-at-tolerance]
+* ['less-or-equal-at-tolerance] <==> ['strictly-less] or ['close-at-tolerance]
+* ['greater-or-equal-at-tolerance] <==> ['strictly-greater] or ['close-at-tolerance]
 
+[note This implies that the exactly one of the following passes: `BOOST_TEST(u < v)`, `BOOST_TEST(u == v)`, `BOOST_TEST(u > v)`.]
```
raffienficiaud: you mean with `BOOST_TEST(rel, tolerance)` right?

doc/testing_tools/testing_floating_points.qbk
```diff
@@ -135,29 +140,42 @@ inequations [link equ2 (2)] and [link equ3 (3)] where all underflow, overflow co
...
+[caution Relation less-at-tolerance is not a Strict Weak Ordering.]
```
raffienficiaud: and cannot be used in eg. `std::map`

doc/examples/tolerance_03.cpp
```diff
@@ -8,29 +8,25 @@
...
-  ratio x (1002, 100); // 10.02
-  ratio y (1001, 100); // 10.01
-  ratio z (1000, 100); // 10.00
-  
-  BOOST_TEST(x == y);  // irrelevant diff by default
-  BOOST_TEST(x == y, tt::tolerance(ratio(1, 2000)));
-  
-  BOOST_TEST(x != z);  // relevant diff by default
-  BOOST_TEST(x != z, tt::tolerance(ratio(2, 1000)));
+  BOOST_TEST(x - d <  x); // less, but not sufficiently
```
raffienficiaud: I would have said "but still too close"

doc/examples/tolerance_03.output
```diff
@@ -1,8 +1,10 @@
 //[example_output
 > tolerance_03
-Running 1 test case...
-test.cpp(23): error: in "test1": check x == y has failed [501/50 != 1001/100]. Relative difference exceeds tolerance [1/1001 > 1/2000]
-test.cpp(26): error: in "test1": check x != z has failed [501/50 == 10/1]. Relative difference is within tolerance [1/501 < 1/500]
+Running 2 test cases...
+test.cpp(18): error: in "failing": check x - d < x has failed [10 - 1.0000000000000001e-005 >= 10]. Relative difference is within tolerance [1e-006 < 0.001]
```
raffienficiaud: I believe that MSVC >= 1900 uses two digits format natively now

akrzemi1:
Ok, I applied most of your suggestions.

I ignored:
> I believe that MSVC >= 1900 uses two digits format natively now (other platforms use that format)

(I do not understand what I am expected to do here)

> Are we required to type `T(0)` ? ...

(This will be addressed separately, but the docs right now say the truth about the implementation)

raffienficiaud:
By the comment "I believe that MSVC >= 1900 uses two digits format natively now", I was saying that MSVC is displaying something like "1.0000000000000001e-05" and not "1.0000000000000001e-005" (05 vs 005)

akrzemi1:
Is the PR acceptable in the current form?

doc/testing_tools/testing_floating_points.qbk
raffienficiaud: I would say "instead we would like to compare u, v and epsilon on the same scale".

doc/testing_tools/testing_floating_points.qbk
```diff
@@ -135,29 +140,42 @@ inequations [link equ2 (2)] and [link equ3 (3)] where all underflow, overflow co
...
+This way all underflow and overflow conditions can be guarded safely. The above however, will not work when `v` or `u` is zero. In such cases the solution is to resort to [link equ1 (1)]. 
+
 [h3 Implementation]
 
-Checks based on equations [link equ4 (4)] and [link equ5 (5)] is implemented by the binary predicate 
-[classref boost::math::fpc::close_at_tolerance `close_at_tolerance`]. The method for comparison (Equ [link equ4 (4)] or [link equ5 (5)]) as well as the
-epsilon are selected at construction. This predicate reports the tolerance to set to make the test successful.
+Binary predicate [classref boost::math::fpc::close_at_tolerance `close_at_tolerance`] implements checks [link equ4 (4)] and [link equ5 (5)].
+The tolerance and the selection between [link equ4 (4)] and [link equ5 (5)] is determined during construction. Apart from the Boolean result,
+the predicate can also return the minimum tolerance to set to make the test successful.
```
raffienficiaud: This is not the "minimum", this is the tolerance. Minimum depends on the comparison that is in use, for close_at_tolerance it would be the minimum, for !close_at_tolerance it would be the maximum. I would just drop the minimum word here.

doc/testing_tools/testing_floating_points.qbk
```diff
@@ -135,29 +140,42 @@ inequations [link equ2 (2)] and [link equ3 (3)] where all underflow, overflow co
...
+
+Unary predicate [classref boost::math::fpc::small_with_tolerance `small_with_tolerance`] implements check [link equ1 (1)].
+
+Assertion __BOOST_TEST__ (when comparing floating-point numbers) uses the following algorithm:
+
+* When either `u` or `v` is zero, evaluates [classref boost::math::fpc::small_with_tolerance `small_with_tolerance`] on the other.
+* When the specified tolerance is zero, preforms direct (native) comparison between `u` and `v`.
+* Otherwise, performs [classref boost::math::fpc::close_at_tolerance `close_at_tolerance`] using [link equ4 (4)].
```
raffienficiaud: This is only for ==, for !=, it uses the negated version of close_at_tolerance (involving with Equ5).

doc/testing_tools/testing_floating_points.qbk
```diff
@@ -83,23 +83,29 @@ BOOST_AUTO_TEST_CASE(test, * utf::tolerance(0.02))
...
 
+[bt_example tolerance_03..tolerance applied in more complex expressions]  
 
-[section:floating_points_comparison_theory Theory]
+[/############################################################################################]
 
-In most cases it is unreasonable to use an `operator==(...)` for a floating-point values equality check.
-The simple, absolute value comparison based, solution for a floating-point values `u`,
-`v` and a tolerance `epsilon`:
+[section:floating_points_comparison_theory Tolerance-based algorithms]
```
raffienficiaud: I am mitigated here. I liked the "Theory" name better here, I would make another section "Practice" or  "Design in the __UTF__" or "Implementation design", containing the "Implementation part", but not contained in this section and visible in the TOC

raffienficiaud:
Thanks, I made some additional remarks. Also please use 2 digits exponents.
I do not know if you are doing that already, but please use new commits (I can only see one commit). It is hard to follow otherwise.

akrzemi1:
About the three-digit vs. two-digit exponent display: the output that is enclosed in the examples, it the output I get when I am running the example programs (I am using MinGW on Windows with GCC 4.8.2). I believe it is the right approach, to give the real life output.

The way I understand your suggestion is that I should manually tweak the results, so that they show something else than what I am getting. I may have misunderstood your suggestion, though?

raffienficiaud:
Yes, as the output that you have is different from mine. When I run it, I have the following (clang/darwin):

```
../doc/examples/tolerance_03.cpp:26: error: in "failing": check x - d < x has failed [10 - 9.9999999999999995e-07 >= 10]. Relative difference is within tolerance [1e-07 < 0.0001]
../doc/examples/tolerance_03.cpp:27: error: in "failing": check x - d != x has failed [10 - 9.9999999999999995e-07 == 10]. Relative difference is within tolerance [1e-07 < 0.0001]
../doc/examples/tolerance_03.cpp:29: error: in "failing": check d > .0 has failed [9.9999999999999995e-07 <= 0]. Absolute value is within tolerance [|1e-06| < 0.0001]
../doc/examples/tolerance_03.cpp:30: error: in "failing": check d < .0 has failed [9.9999999999999995e-07 >= 0]
```

If the compilers do not agree, I would go for the unified solution. Either 2 or 3 digits, but all floating point representation should be the same. I believe we are using 2 digits now, also in the unit tests.

akrzemi1:
> I am mitigated here. I liked the "Theory" name better here, I would make another section "Practice" or "Design in the UTF" or "Implementation design", containing the "Implementation part", but not contained in this section and visible in the TOC

Perhaps we can reach some compromise. Can you tell me what you need section "Theory" for and what you want to communicate with it? I found it confusing that it described how components like  `close_at_tolerance` are implemented (e.g., what they return and how they are initialized).

The current state proposed with this PR is not clean either. My goal, however, is to clearly and in short, describe what is actually checked when I type `BOOST_TEST(v == u, tt:tolerance(eps))`. There is some overlap, but perhaps this needs two pages: one saying what is computed, the other saying why? (This why is what i would call the "theory", but it could also be the "rationale", explaining why you prefer to implement it this way or the other, and I do not necessarily see it in the section "Writing unit tests".

Please tell me what you think about this.

raffienficiaud:
'design rational' is ok as well (but I prefer theory better), and you are right in saying that it is confusing to describe `close_at_tolerance` there. Also, `close_at_tolerance` is part of a public but not readily accessible API.

In this "Theory", I see everything related to the equations and the references. I think it is nice to have it directly in the documentation. Putting it too far from the topic would make it less accessible as well: the idea is to help ppl writing better tests with UTF and floating points. What about referencing theory in the "floating point comparisons" part, putting the section "theory" as a final subsection of it, and removing everything related to the code from this part?

A possible TOC would be:
```
x- floating point comparison
  x.1 - Enabling tolerance for user-defined types
  x.2 - Design rational (where close_at_tolerance is described)
  x.3 - Theory behind floating point comparisons
```

Would that be better?

akrzemi1:
> x- floating point comparison
  x.1 - Enabling tolerance for user-defined types
  x.2 - Design rational (where close_at_tolerance is described)
  x.3 - Theory behind floating point comparisons

Where do I put the equations: in x.2 or x.3? Also, I need a place to say "`BOOST_TEST(v == u, tt:tolerance(eps))` uses 'this comparison'" -- where should it go?

Can I change it to:

x- floating point comparison
  x.1 - Enabling tolerance for user-defined types
  x.2 - Tolerance-based comparison (no mention of close_at_tolerance)
  x.3 - Theory behind floating point comparisons

akrzemi1:
Ok, I added a new commit. It should address all the remarks.

raffienficiaud:
Thanks

raffienficiaud:
I just checked and the reported tolerance by close_at_tolerance is described in the reference documentation. Should be enough.

PR #74, opened 2015-06-02
testing: run tests conditionally

akrzemi1:
As an illustration, I configured three examples from documentation, so that they are only compiled when support for variadic macros is present on the compiler. I am using a feature from Boost.Config: http://www.boost.org/doc/libs/1_58_0/libs/config/doc/html/boost_config/build_config.html
It uses predefined platform macros (rather than tester name) to detect the availability of the feature.


raffienficiaud:
Super, many thanks!

PR #85, opened 2015-08-30
Fix #11582, and other enhancements.

BenWiederhake:
I ran all regression tests, and see no error on my machine.

This pull requests handles several issues. I don't know how you prefer it, so I can split this pull requests into multiple small chunks if you wish.

The first two commits: Fixes #11582 on trac. These commits provide a starting point for discussion. I'm willing to work together on a better solution.

Commits three and four care about typos, I don't expect any problems here. This is a strict superset of the other pull request.

The last commit fixes current issues in <boost/test/detail/throw_exception.hpp>.
As far as I can see, you are currently working on this and already have fixed in your branch "new_cla"; however you haven't made it available yet. Since this change is atomic and easy to separate, I suggest applying it now instead of waiting on new_cla. YMMV.

EDIT: Github wrongly displays the order of commits in the overview below (huh?)
In my PR, *first* add the tests, *then* provide the fix. Again, YMMV, and I'm willing to change the order if you wish.

### DIFF
--- a/include/boost/test/data/monomorphic/generators/random.hpp
+++ b/include/boost/test/data/monomorphic/generators/random.hpp
@@ -36,7 +36,7 @@ namespace data {
 
 namespace {
 nfp::keyword<struct seed_t>         seed;
-nfp::keyword<struct distrbution_t>  distribution;
+nfp::keyword<struct distribution_t> distribution;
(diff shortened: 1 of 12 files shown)


rogeeff:
BOOST_NORETURN and typo in in develop

PR #86, opened 2015-09-06
Fix broken support for building with -fno-exceptions on GCC.

jzmaddock:
Boost.Test used to support building with -fno-exceptions on GCC, this patch reinstates that support.


rogeeff:
thx for the patch

PR #87, opened 2015-09-18
Fix compilation with BOOST_NO_EXCEPTIONS

MarcelRaad:
The return type of throw_exception was missing.

### DIFF
--- a/include/boost/test/detail/throw_exception.hpp
+++ b/include/boost/test/detail/throw_exception.hpp
@@ -32,7 +32,7 @@ namespace ut_detail {
 #ifdef BOOST_NO_EXCEPTIONS
 
 template<typename E>
-inline
+inline void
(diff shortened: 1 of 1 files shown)


raffienficiaud:
Thanks,

is it `inline void BOOST_NO_RETURN` or `BOOST_NO_RETURN inline void` ?

MarcelRaad:
With both MSVC14 and clang-cl, both versions work.

raffienficiaud:
Applied, thanks.

I changed a bit to be in accordance with the documentation I found:
* http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1453.htm
* https://msdn.microsoft.com/en-us/library/k6ktzx3s.aspx

MarcelRaad:
Great, thanks!

PR #94, opened 2016-02-01
fix a typo

rick68:
fix typo, rename `BOOST_TEST_I_ASSRT` to `BOOST_TEST_I_ASSERT`


raffienficiaud:
Hi,
This is an internal interface, why would your changes make any difference?

rick68:
I though that is a typo, but it appears not.

PR #96, opened 2016-03-27
Fix for clang on windows

jzmaddock:
When I try to build with clang on windows (4.2.1 Compatible Clang 3.9.0 (trunk), on top of VC14) I get an error complaining that accessing "this" within an __except block is unimplemented.  This patch just stores the value in a local variable to make it available with the __finally clause.

Allows Boost.Test to be built with clang on windows, in VC compatibility mode.


raffienficiaud:
Applied to topic/PR96-clang-windows-fix, will be merged to develop in a couple of hours (internal CI).

Thanks

raffienficiaud:
Merged to develop

PR #99, opened 2016-04-19
Fix crashes with RTTI-off on MSVC

MarcelRaad:
When RTTI is off on MSVC, BOOST_NO_TYPEID is not defined,
but only static typeid works. Dynamic typeid crashes at runtime.


raffienficiaud:
Thanks, currently under testing. Would you mind creating a ticket on trac?

MarcelRaad:
Thanks! I've created a ticket here:
https://svn.boost.org/trac/boost/ticket/12224

raffienficiaud:
Thanks, under testing, should be in 1.62. I am closing the PR.

PR #100, opened 2016-05-19
Allow joins to properly join grids

Houndie:
Joining two grids (of the same arity) would formerly cause a compiler
error.

See Trac #12216

### DIFF
--- a/include/boost/test/data/for_each_sample.hpp
+++ b/include/boost/test/data/for_each_sample.hpp
@@ -63,7 +63,7 @@ invoke_action( Action const& action, T&& args, std::true_type /* is_tuple */ )
 {
     invoke_action_impl( action,
                         std::forward<T>(args),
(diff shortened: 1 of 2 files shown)


Houndie:
Note that this pull request uses some c++14 features...I saw c++14 features used in grid.hpp (with auto return type) so I went ahead and used them myself.  If these changes are in places that support older standards, this pull request won't work as is...just let me know.

Houndie:
This patch apparently causes some segfaults when running test cases compiled *without* optimization.  I'll leave this patch up for inspirational purposes, but it might not work as is.

Houndie:
Patch has been updated to no longer cause segmentation faults.  It still should only be considered "lightly tested".

raffienficiaud:
Unfortunately your changes require C++14 and are breaking some other tests. So it will take some additional time to get it right.

raffienficiaud:
Fixed by rev 2fc875b9b22c717e17e56171501bd9aa3051213c

PR #103, opened 2016-09-16
Syntactic change silences latest gcc warnings

eldiener:
Gcc issues warnings whenever it encounters the expansion of the macro, claiming the 'if' statements could be confusing. This purely syntactical change of removing the 'if' one liners in favor of the 'if' block silences the warnings.


raffienficiaud:
Merged to PR/remove-gcc6-warnings

raffienficiaud:
In develop

PR #104, opened 2016-10-16
Fix gcc's warning/error when building with -Wmisleading-indentation

Romain-Geissler-1A:
Hi,

I am having troubles when including "boost/test/unit_test.hpp" with gcc 6 and -Wall -Werror turned on. gcc (wrongly) complains about misleading indentation in the macro DEFINE_FPV_COMPARISON. This is likely caused by the fact that we have extra indentation after each line so that the backlslashes are properly aligned, and the macro is eventually putting this on a single line. Anyway, an easy fix is to use braces around all block of code.

Ok for the merge ?

(Similar to https://github.com/boostorg/spirit/pull/222)

Cheers,
Romain

### DIFF
--- a/include/boost/test/tools/fpc_op.hpp
+++ b/include/boost/test/tools/fpc_op.hpp
@@ -178,16 +178,22 @@ public:                                                                 \
     eval( Lhs const& lhs, Rhs const& rhs )                              \
     {                                                                   \
         if( lhs == 0 )                                                  \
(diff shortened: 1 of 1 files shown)


raffienficiaud:
Thanks, this is a duplicate of #103 that is already merged to a feature/check branch.

raffienficiaud:
In develop

PR #108, opened 2017-01-18
Change capital variable names to lowercase

AdamMajer:
Capitals clash by convention with #define, and in this case
VERSION clashes with GNU Autotools. This means projecting using
GNU Autotools and Boost::Test library no longer compile.

To fix this problem and possibly others down the line,
downcase all the string constants in the rt::runtime_config
namespace. (trac 12748)

### DIFF
--- a/doc/examples/example51.run.cpp
+++ b/doc/examples/example51.run.cpp
@@ -13,7 +13,7 @@ using namespace boost::unit_test;
 
 BOOST_AUTO_TEST_CASE( test_case0 )
 {
-  if( runtime_config::get<log_level>( runtime_config::LOG_LEVEL ) < log_warnings )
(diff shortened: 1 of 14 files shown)


AdamMajer:
As to obvious "why not just change VERSION to something else?" - so it's consistent.

raffienficiaud:
Thanks for the PR. While I understand for the `VERSION` macro clash, I would like to keep this minimal. So I would rather rework your PR for fixing this issue and making those variables/macros more boost.test specific.

raffienficiaud:
Merged to branch `origin/topic/12748-VERSION-macro-clashes`, thanks

AdamMajer:
Thanks fine too. Thank you.

raffienficiaud:
Merged to develop, closing.

PR #112, opened 2017-02-14
Deliberate-failure tests shouldn't be optimized

aargor:
A few tests designed to deliberately fail appeal to UB (undefined
behavior), but as the MSVC optimizer improves we will take advantage of
that to remove UB statements. In order for deliberate failures to occur,
the tests must have the optimizer turned off.

### DIFF
--- a/doc/examples/example22.run-fail.cpp
+++ b/doc/examples/example22.run-fail.cpp
@@ -23,8 +23,8 @@ void goo( int )
 {
 }
 
-#if defined(BOOST_MSVC)
-// compiler optimizations may cause this code NOT to crash.
+#if defined(BOOST_MSVC) && (BOOST_MSVC > 1900)
+// VS2017+ compiler optimizations may cause this code NOT to crash.
 #pragma optimize("", off)
 #endif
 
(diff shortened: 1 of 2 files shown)


aargor:
Hello, I'm from the MSVC++ compiler team. I created the above pull request because as we're improving the optimizer, we've tested internally that a few tests will fail as we more successfully optimize in the presence of undefined behavior. The team I'm on expects that these tests will fail with the soon-to-be released VS2017, because we optimize away the code that's designed to trigger the faults.

I hope this is the right way to contribute back to the Boost project. Please let me know if there's anything else I should do.

Thanks,
Aaron Gorenstein
Microsoft Visual C++

raffienficiaud:
Hi,

Since you released VS2017, is this PR still relevant? If yes, I would welcome a version of the compiler be used together with the `BOOST_MSVC` macro.

aargor:
Hi,

Yes, I've just confirmed that the patch is needed with the latest compiler version. Thanks for the feedback! I think I am now using the BOOST_MSVC macro as you requested (and I've confirmed this still addresses the reported issues), but let me know if you were thinking of something else.

(For concreteness, in VS2017 _MSC_VER is 1910, and VS2015 it's 1900. My understanding is that BOOST_MSVC inherits that value.)

Thanks again,
Aaron

raffienficiaud:
In the test queue, thanks

raffienficiaud:
Merged to develop, thanks.

PR #114, opened 2017-03-08
Silence unreachable code warning in MSVC (/W4)

norbertwenzel:
Disable level 4 warning C4702 on MSVC. Since the comments indicate, that the unreachable `return 0` silences some other compiler warning I think it is best to simply ignore the MSVC warning.

### DIFF
--- a/include/boost/test/impl/execution_monitor.ipp
+++ b/include/boost/test/impl/execution_monitor.ipp
@@ -1287,7 +1287,7 @@ execution_monitor::execute( boost::function<int ()> const& F )
 
 #endif // !BOOST_NO_EXCEPTIONS
 
-    return 0;  // never reached; supplied to quiet compiler warnings
+    BOOST_UNREACHABLE_RETURN(0);  // never reached; supplied to quiet compiler warnings
 } // execute
 
 //____________________________________________________________________________//
--- a/include/boost/test/utils/named_params.hpp
+++ b/include/boost/test/utils/named_params.hpp
@@ -131,23 +131,23 @@ struct nil {
 #else
     operator T const&() const
 #endif
-    { nfp_detail::report_access_to_invalid_parameter(true); static T* v = 0; return *v; }
+    { nfp_detail::report_access_to_invalid_parameter(true); static T* v = 0; BOOST_UNREACHABLE_RETURN(*v); }
 
     template<typename T>
     T any_cast() const
-    { nfp_detail::report_access_to_invalid_parameter(true); static typename remove_reference<T>::type* v = 0; return *v; }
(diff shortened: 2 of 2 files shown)


raffienficiaud:
Hi, have you considered using `BOOST_UNREACHABLE_RETURN` ? (see [there](http://www.boost.org/doc/libs/1_64_0/libs/config/doc/html/boost_config/boost_macro_reference.html#boost_config.boost_macro_reference.boost_helper_macros)) If not, would you please give a try?

norbertwenzel:
I did try `BOOST_UNREACHABLE_RETURN` and it works for Release builds only. It seems that MSVC Debug builds require these return statements they warn about in Release builds.

I've tested VS2017 (15.2) with Boost 1.64 which does *not* trigger a warning in `execution_monitor.ipp` as I intended to fix. But it triggers the same warning in a different spot, `util/named_params.hpp:138`, which also can be fixed using `BOOST_UNREACHABLE_RETURN` only in Release builds. Debug builds cause level 1 [compiler warning C4716](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4716), which is by default treated as a build error.

I've also tested VS2015 (14.x) with Boost 1.63, which led to this PR and, if I remember correctly (I don't have that computer available at the moment), produced the exact same problems as described above. The macro works in Release builds only and causes actual compile errors in Debug.

In Release this macro works fine and is definitely a superior solution to my `pragma push/pop` PR.

raffienficiaud:
What about this: provide me with the commit, and I'll try to run it on other platforms. If there is something wrong with the macro, then I'll report to the boost ML.

norbertwenzel:
You mean the commit using the macro? Sure. Should I just update the one I fixed in the initial PR or should I also take care of the ones in `named_params.hpp`?

norbertwenzel:
I've added both changes in different commits. so you can separate them if necessary. 
The code is tested and compiles flawlessly in MSVC 2017 at /W4 in Release builds. **Debug builds fail as described before.**

norbertwenzel:
I've also tested with MVSC 2015 now. The behaviour is the same as in 2017. Release build is fine, but Debug build fails.

raffienficiaud:
On branch ``topic/PR114-unreachable-code-warning-MSVC``, tests running

jeking3:
Perhaps this never got merged.  Is it the same issue as #180 ?

raffienficiaud:
Yes, the PR as proposed causes build issues and I never got the time to rework it.

raffienficiaud:
I am not sure why you were seeing warnings, I believe the `BOOST_UNREACHABLE_RETURN` in `execution_monitor.ipp` is legit, but the ones in `named_params.hpp` are not. I am keeping the one in `execution_monitor.ipp` to finally close this PR.

raffienficiaud:
Merged to master

PR #115, opened 2017-04-07
WIP: use std::function if c++11 <functional> is available

jhunold:
All except one test green. 
`test-tree-management-test` is failing as somehow the conversion of `std::function` to `bool` is not working as expected in 
`BOOST_TEST( tc1->p_test_func );` in line 118
Tested c++03 compatibility with gcc-6.2 in c++03 mode

### DIFF
--- a/doc/examples/example08.run-fail.cpp
+++ b/doc/examples/example08.run-fail.cpp
@@ -27,7 +27,11 @@ bool init_unit_test()
 {
   double params[] = { 1., 1.1, 1.01, 1.001, 1.0001 };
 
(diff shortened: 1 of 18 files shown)


raffienficiaud:
Hi, thanks for the PR. I fail to understand why should we move to `std::function`, is there any rationale? doesn't `boost::function` do the job?

jhunold:
Current clang is starting to warn about deprecated c++03/c++98 features used in the implementation ```boost::function```. The easiest way to disable them is to simply switch to ```std::function``` if it is available. And I think that using the ```std::``` where possible is the correct long term solution.

PR #119, opened 2017-09-12
Prevent some warnings

bebuch:


include/boost/test/impl/framework.ipp
```diff
@@ -1568,6 +1568,9 @@ run( test_unit_id id, bool continue_test )
             break;
         case 1:
             seed = static_cast<unsigned>( std::rand() ^ std::time( 0 ) ); // better init using std::rand() ^ ...
+#if __has_cpp_attribute(fallthrough)
```
raffienficiaud: I believe this is not supported by all compilers, especially C++03

raffienficiaud:
I think your changes are not C++03 compliant, and the `__has_cpp_attribute` is not standard (at least from what I read [here](http://en.cppreference.com/w/cpp/experimental/feature_test)). 

Would you like to rework your patch, or should I close the PR?

raffienficiaud:
Seems to duplicate #122 , what do you think

PR #121, opened 2017-11-09
fix compiler warning

jeking3:
I'm not certain if this is correct - but "l" isn't used currently.


Romain-Geissler-1A:
Ping @raffienficiaud

raffienficiaud:
Thanks, on the tests queue

PR #127, opened 2018-01-31
Silence 'unused variable' warning

mclow:
Mac OS doesn't fall into the "UNIX based debugging" or "windows based debugging", so we get the "do nothing" behavior.


raffienficiaud:
Thanks for the patch, in ``next``.

PR #130, opened 2018-02-13
Add option to wait for keypress before application terminates

ShadowIce:
I've added a command line parameter to pause execution until a key is pressed. This should be very useful if you run tests directly in Visual Studio with attached debugger because Visual Studio just closes the console window after the application terminates.


raffienficiaud:
Hi,

Why are you closing this?

ShadowIce:
Because Microsoft added a feature to keep the console window open after the process terminates, so from my point of view this is not necessary any more (https://visualstudio.uservoice.com/forums/121579-visual-studio-ide/suggestions/31124656-toggle-console-window-to-stay-open-on-off-settin).

But if you or someone else still needs this feature I can reopen the PR again.

PR #136, opened 2018-03-26
Change Windows.h include to all-lowercase.

rdoeffinger:
Otherwise cross-compiling with MinGW on Linux
will fail.


raffienficiaud:
In branch `PR/136-Windows-h-all-lowercase`, tests cycling.

PR #142, opened 2018-04-26
Make under_debugger work on apple (iOS/macOS)

thughes:
Simply uses the documented way from Apple to detect the debugger: https://developer.apple.com/library/content/qa/qa1361/_index.html


raffienficiaud:
There is also this: https://zgcoder.net/ramblings/osx-debugger-detection.html and the approach in the PR is said to be for debug builds only.

raffienficiaud:
In master, thanks!

PR #151, opened 2018-08-24
Fix warning: BOOST_CLANG is not defined, evaluates to 0

mloskot:
The diagnostic is issued when compiled with GCC/clang flag `-Wundef`.

Although in case `BOOST_CLANG` is not defined, ie. is not a macro and `BOOST_CLANG` identifier is considered to be zero, the `#ifdef` directive makes the intention clearer.

### DIFF
--- a/include/boost/test/detail/enable_warnings.hpp
+++ b/include/boost/test/detail/enable_warnings.hpp
@@ -26,7 +26,7 @@
 # pragma warning(pop)
 #endif
 
-#if BOOST_CLANG
+#ifdef BOOST_CLANG
 #pragma clang diagnostic pop
 #endif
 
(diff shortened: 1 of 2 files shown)


raffienficiaud:
Wouldn't 

```
#if defined(BOOST_CLANG) && (BOOST_CLANG == 1)
```

be safer? I updated the branch `topic/PR-151-BOOST_CLANG-define` accordingly.

mloskot:
Or, `BOOST_CLANG != 0`. Sure, why not.

raffienficiaud:
Merged to next.

raffienficiaud:
Thanks, merged to develop

mloskot:
Thanks!

Issue #149, opened 2018-08-21
Setting color_output=no does not disable the output of color format codes


Markus87:
I think --color_output=no should not just use the default color for everything, it should stop color format codes from being outputted completely. The current behaviour is very misleading to me.
https://github.com/boostorg/test/blob/35c9acf5bd38d625ebd45cf940c5b9a19aa2295b/include/boost/test/utils/setcolor.hpp#L304

To understand why it is a problem let me describe my test setup:
I use ctest to run boost tests and instructed ctest to output everything to a file.
The file contains all color codes that boost writes, --color_output=yes/no only changes if its all the same color or not, the codes remain any way.

toonetown:
Agreed.  This also happens when trying to run unit tests within Xcode's debugger windows (which doesn't support ANSI colors).

I was able to work around it (in my particular case) by tweaking the `#define BOOST_TEST_SCOPE_SETCOLOR`.  I added these lines before including my main unit test headers:

```
#include <boost/test/utils/setcolor.hpp>
#undef BOOST_TEST_SCOPE_SETCOLOR
#define BOOST_TEST_SCOPE_SETCOLOR(is_color_output, os, attr, color)
```

That basically just disables ansi color altogether.

toonetown:
IMO, I think that if you say `--enable_color=no`, it should completely skip the ansi codes altogether.

toonetown:
A proposed patch can be found at https://github.com/boostorg/test/pull/154

raffienficiaud:
Closed through PR #154

PR #154, opened 2018-08-30
When specifying --color_output=no, don't output color codes

toonetown:
As described in https://github.com/boostorg/test/issues/149, specifying `--color_output=no` still will output the color codes, it just outputs the codes for "normal" output.

A better approach would be if the user disables color output completely, the codes should not be included at all.  This PR attempts to do that.


raffienficiaud:
Thanks, merged to develop.

PR #166, opened 2018-09-18
Fix some of the visibility issues

apolukhin:


apolukhin:
Do not forget to put `visibility=global` [here](https://github.com/boostorg/test/blob/4df5b095eaa967f5e2687978bd3d9467b758f28e/example/Jamfile.v2#L18) and [here](https://github.com/boostorg/test/blob/4df5b095eaa967f5e2687978bd3d9467b758f28e/test/Jamfile.v2#L18)

Or force the shared library singleton instances usage by some code magic, like `singleton<type>::instance().variable`.

raffienficiaud:
Superseded by rev 49cd9a17847aa24a925749a3d39ffc045e9fb761 (getting rid of the broken singleton class).

Thanks for your support with this issue!

PR #172, opened 2018-10-19
Check for non-used variables when NDEBUG is defined

github-peter:
When NDEBUG is defined, the compiler may warn about variables not used. They are only used in assert's.


raffienficiaud:
I modified a bit your contribution (see 5a2ee25b5445a6ef411f98f6aa22ceaa63efdce2). Currently in next.
Thanks

raffienficiaud:
Merged to develop, thanks.

PR #190, opened 2018-11-23
fix warning on gcc-7.3 in cygwin claiming master_test_suite is declared differently

jeking3:
See an example here:
https://ci.appveyor.com/project/jeking3/thread/builds/20518379/job/m0ppxtti4invy0ka?fullLog=true#L1459

```
gcc.compile.c++ bin.v2\libs\thread\test\test_thread_exit.test\gcc-7.3.0\rls\cxstd-03-iso\trgt-os-cygwn\thrdp-pthrd\thrd-mlt\vsblt-hdn\test_thread_exit.o
In file included from ./boost/test/tree/auto_registration.hpp:21:0,
                 from ./boost/test/unit_test_suite.hpp:17,
                 from ./boost/test/unit_test.hpp:19,
                 from libs\thread\test\test_thread_exit.cpp:17:
./boost/test/tree/test_unit.hpp:232:33: warning: 'boost::unit_test::master_test_suite_t& boost::unit_test::framework::master_test_suite()' redeclared without dllimport attribute: previous dllimport ignored [-Wattributes]
     friend master_test_suite_t& boost::unit_test::framework::master_test_suite();
                                 ^~~~~
```

PR #195, opened 2019-01-08
Fix MinGW compilation problems

Lastique:
This fixes problems with MinGW gcc 5.3 that were discovered in Boost.Log CI [builds](https://ci.appveyor.com/project/Lastique/log/build/job/v59mk63a0swtdic1):

- `master_test_suite` function redeclared without `dllimport`
- `vsnprintf` not declared

The latter problem I couldn't reproduce locally, but I guess it is caused by a missing include of stdio.h. If that doesn't help, another fix might be needed.

### DIFF
--- a/include/boost/test/impl/execution_monitor.ipp
+++ b/include/boost/test/impl/execution_monitor.ipp
@@ -48,7 +48,9 @@
 #include <cassert>              // for assert
 #include <cstddef>              // for NULL
 #include <cstdio>               // for vsnprintf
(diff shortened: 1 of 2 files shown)


Lastique:
The `dllimport` problem is also fixed by https://github.com/boostorg/test/pull/190.

raffienficiaud:
In next

raffienficiaud:
Merged to develop, thanks

PR #185, opened 2018-11-11
Added CI framework

jeking3:
- travis with valgrind, cppcheck, ubsan, codecov, covscan (future)
  - appveyor with MSVC 2010 through 2017, cygwin 32/64, mingw 32/64
  - README, LICENSE, etc.
  - Fixed example Jamfile to build examples based on C++ features

Please see my fork's PR for build results.
https://github.com/jeking3/test/pull/1

A number of builds are failing for various reasons - mostly clang due to release optimizer issues in doc_example22 and doc_example23 that needs to be resolved.

As we also discussed by email, the static and UBSAN analyses may need additional work to annotate the intentional issues (this repository is unique, as we're testing a test library).  I can update this PR to mark those as expected failures once we resolve the others.

PR #205, opened 2019-02-10
Fix MinGW vsnprintf compile errors and warnings

Lastique:
This PR:

- Adds Appveyor CI config file.
- Fixes compile errors on MinGW about missing declaration of `vsnprintf`. The attempted fix in https://github.com/boostorg/test/pull/195, although correct, did not fix the error.
- Silences gcc warnings about unused global variables.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,112 @@
+# Copyright 2019 Andrey Semashev
+# Distributed under the Boost Software License, Version 1.0.
+# (See accompanying file LICENSE_1_0.txt or copy at http://boost.org/LICENSE_1_0.txt)
+
+version: 1.0.{build}-{branch}
+
+shallow_clone: true
+
+branches:
+  only:
+    - master
+    - develop
+    - /topic\/.*/
+
+environment:
+  matrix:
+# AppVeyor doesn't provide 64-bit compilers for these MSVC versions
+#    - TOOLSET: msvc-9.0
+#      ADDRESS_MODEL: 64
+#      APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
+#    - TOOLSET: msvc-10.0
+#      ADDRESS_MODEL: 64
+#      APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
+#    - TOOLSET: msvc-11.0
+#      ADDRESS_MODEL: 64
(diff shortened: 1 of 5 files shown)


raffienficiaud:
There is already a PR about AppVeyor, see #185 . For some reason it is not active yet. It would be good if you can split the PR on those orthogonal issues. Thanks!

Lastique:
That PR sets up MinGW differently than my config, I'm not sure it is equivalent. Also, it doesn't enable CI for `topic` branches. I've included the config because I had to ensure the problem reproduces and test the change in Appveyor since I couldn't reproduce it locally. 

The config is in a separate commit, you can skip merging it. If you want, I can remove this commit from this PR. Please, confirm.

> For some reason it is not active yet. It would be good if you can split the PR on those orthogonal issues.

You have to add the GitHub project to your Appveyor project list in order to get it running. You should have admin rights for the GitHub project though.

raffienficiaud:
I split the PR in 3 branches:

Concerning the `vsnprintf` issue (that is the purpose of the PR), it is currently running the tests in the branch `topic/PR-205-Fix-MinGW-vsnprintf-compile-errors-warnings`. I'll merge that once the tests have cycled.

The GCC warnings is on branch `topic/silence-gcc-warnings` and the AppVeyor one is in the branch `topic/CI-appveyor`. I need to merge the Appveyor work with the other running PR on CI. Concerning the gcc warning, I am not sure: I can see some of the warnings, but all those symbols are used at some point. Since I do not know what `BOOST_ATTRIBUTE_UNUSED` does and since this is just about warnings (lesser priority), I suggest we wait a bit. The other option for silencing the warnings would be to add the proper pragmas in the `suppress_warnings.hpp` file.

Lastique:
> Since I do not know what `BOOST_ATTRIBUTE_UNUSED` does...

This is [`__attribute__((unused))`](https://gcc.gnu.org/onlinedocs/gcc-8.2.0/gcc/Common-Variable-Attributes.html#index-unused-variable-attribute) for gcc, which explicitly indicates that the variable may appear unused in the program. Disabling the warning with a pragma may work, but it may also disable the warning for genuinely leftover unused variables, so I prefer the attribute.

raffienficiaud:
Changes merged to next except for the AppVeyor (tried to enable the thing from Appveyor but still waiting for some approval by boost.org).

Closing, thanks!

PR #211, opened 2019-03-19
'windows.h' should be lower case.

rainerdeyke:
When cross-compiling for Windows on Linux using mingw, <window.h> should be all lower case.


raffienficiaud:
Currently building on branch `topic/PR-211-windows-h-lower-case`

raffienficiaud:
In develop, thanks

raffienficiaud:
In master.

PR #227, opened 2019-06-14
Add printf format checking attribute to report_error

ecatmur:
On gcc and clang, add `__attribute__((__format__))` checking to the `report_error` function.

Note: clang warns `-Wformat-nonliteral` (off by default, but good practice) when a format string argument not annotated by `__attribute__((__format__))` is passed to `vprintf` etc.; gcc does not. See https://clang.llvm.org/docs/AttributeReference.html#format .

Plus fixes for bugs exposed by this change:
* Cast faulting addresses to `uintptr_t` for formatting as `0x%08lx`. If there is an LLP64 platform that uses Posix signals (i.e. not Win64, which uses SEH) this should be changed to use `PRIxPTR` format specifier.
* Fix swapped `si_code`/`si_addr` (& `si_band`) format arguments.
* Add missing `%s` to format diagnostic information.

### DIFF
--- a/include/boost/test/impl/execution_monitor.ipp
+++ b/include/boost/test/impl/execution_monitor.ipp
@@ -242,7 +242,7 @@ extract( boost::exception const* ex )
 //____________________________________________________________________________//
 
 static void
-#ifdef __GNUC__
+#if __GNUC__ >= 3
 __attribute__((__format__ (__printf__, 3, 0)))
 #endif
 report_error( execution_exception::error_code ec, boost::exception const* be, char const* format, va_list* args )
@@ -263,7 +263,7 @@ report_error( execution_exception::error_code ec, boost::exception const* be, ch
 //____________________________________________________________________________//
 
 static void
-#ifdef __GNUC__
+#if __GNUC__ >= 3
 __attribute__((__format__ (__printf__, 3, 4)))
 #endif
 report_error( execution_exception::error_code ec, boost::exception const* be, char const* format, ... )
@@ -279,7 +279,7 @@ report_error( execution_exception::error_code ec, boost::exception const* be, ch
 //____________________________________________________________________________//
 
 static void
-#ifdef __GNUC__
+#if __GNUC__ >= 3
 __attribute__((__format__ (__printf__, 2, 3)))
 #endif
 report_error( execution_exception::error_code ec, char const* format, ... )
--- a/include/boost/test/impl/execution_monitor.ipp
+++ b/include/boost/test/impl/execution_monitor.ipp
@@ -242,6 +242,9 @@ extract( boost::exception const* ex )
 //____________________________________________________________________________//
 
 static void
+#ifdef __GNUC__
+__attribute__((__format__ (__printf__, 3, 0)))
+#endif
 report_error( execution_exception::error_code ec, boost::exception const* be, char const* format, va_list* args )
 {
     static const int REPORT_ERROR_BUFFER_SIZE = 4096;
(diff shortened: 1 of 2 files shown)


include/boost/test/impl/execution_monitor.ipp
```diff
@@ -1298,7 +1307,7 @@ execution_monitor::execute( boost::function<int ()> const& F )
 #if defined(BOOST_NO_TYPEID) || defined(BOOST_NO_RTTI)
                               "unknown boost::exception" ); }
 #else
-                              boost::diagnostic_information(ex).c_str() ); }
+                              "%s", boost::diagnostic_information(ex).c_str() ); }
```
raffienficiaud: I am not sure to understand why this change would be needed.

include/boost/test/impl/execution_monitor.ipp
```diff
@@ -534,23 +543,23 @@ system_signal_exception::report() const
...
             report_error( execution_exception::system_fatal_error,
                           "memory access violation at address: 0x%08lx: non-existent physical address",
-                          m_sig_info->si_addr );
+                          (uintptr_t) m_sig_info->si_addr );
             break;
         case BUS_OBJERR:
             report_error( execution_exception::system_fatal_error,
                           "memory access violation at address: 0x%08lx: object specific hardware error",
-                          m_sig_info->si_addr );
+                          (uintptr_t) m_sig_info->si_addr );
```
raffienficiaud: `uintptr_t` might be unavailable on some compilers

include/boost/test/impl/execution_monitor.ipp
```diff
@@ -242,6 +242,9 @@ extract( boost::exception const* ex )
 //____________________________________________________________________________//
 
 static void
+#ifdef __GNUC__
+__attribute__((__format__ (__printf__, 3, 0)))
```
raffienficiaud: How can we ensure that this attribute is available for the target compiler?

raffienficiaud:
Hi,

Thanks for the PR, tests are passing. However I have some comments. Would you please take the time to address those?

Thanks

include/boost/test/impl/execution_monitor.ipp
ecatmur: It's a security issue as well as a correctness issue:

gcc.compile.c++ ../../../bin.v2/libs/test/build/gcc-7/debug/link-static/threading-multi/visibility-hidden/execution_monitor.o
In file included from ../../../libs/test/src/execution_monitor.cpp:16:0:
../../../boost/test/impl/execution_monitor.ipp: In member function ‘int boost::execution_monitor::execute(const boost::function<int()>&)’:
../../../boost/test/impl/execution_monitor.ipp:1311:73: error: format not a string literal and no format arguments [-Werror=format-security]
                               boost::diagnostic_information(ex).c_str() ); }
                                                                         ^
cc1plus: some warnings being treated as errors

I'll add a test illustrating this.

include/boost/test/impl/execution_monitor.ipp
ecatmur: You're already using `uintptr_t` in this file: 
https://github.com/boostorg/test/blob/4f3806d80252081bc87e71d97d5b0c8dfd063e17/include/boost/test/impl/execution_monitor.ipp#L1131
https://github.com/boostorg/test/blob/4f3806d80252081bc87e71d97d5b0c8dfd063e17/include/boost/test/impl/execution_monitor.ipp#L87
https://github.com/boostorg/test/blob/4f3806d80252081bc87e71d97d5b0c8dfd063e17/include/boost/test/impl/execution_monitor.ipp#L91

include/boost/test/impl/execution_monitor.ipp
ecatmur: I'll change the check to `__GNUC__ >= 3`. Or would you prefer to set this only for specific compilers (I've tested with gcc 6-8 and clang 5-9)?

raffienficiaud:
Thank you for having addressed all the comments and for the quality of the work. This is currently in next and should be merged to develop pretty soon.

raffienficiaud:
In master, closing. Thanks!

PR #231, opened 2019-10-04
OpenBSD is missing SI_ASYNCIO and SI_MESGQ

kernigh:
Check if macros SI_ASYNCIO and SI_MESGQ are defined as macros.  This
allows to run tests on OpenBSD 6.5, where the macros are missing.

This is identical to [patch-boost_test_impl_execution_monitor_ipp](https://cvsweb.openbsd.org/cgi-bin/cvsweb/ports/devel/boost/patches/patch-boost_test_impl_execution_monitor_ipp?rev=1.3&content-type=text/x-cvsweb-markup) in
OpenBSD Ports, except that I added a comment.


kernigh:
This change assumes that system headers always define SI_ASYNCIO and SI_MESGQ as macros. This seems to be true in at least [FreeBSD](https://svnweb.freebsd.org/base/head/sys/sys/signal.h?revision=352807&view=markup#l408), [glibc](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/bits/siginfo-consts.h;hb=021197483e8fa9999644c96d7dd7b5c3cc95867a#l59), [illumos](https://github.com/illumos/illumos-gate/blob/052042909d20316395a71c237105dffb9ce6c22f/usr/src/uts/common/sys/siginfo.h#L134), and [musl](https://git.musl-libc.org/cgit/musl/tree/include/signal.h#n37).

raffienficiaud:
Thanks for the report, this is currently building in branch `topic/PR-231-defines-missing-openbsd`

raffienficiaud:
Merged to `next`.

raffienficiaud:
In master, closing

PR #232, opened 2019-10-24
fix timeout in windows

mmitti:
Visual studio 2019 + Boost 1.71.0 + Windows 10 64bit

I use boost test with timeout(240) like below, it will timeout error.

```cpp
BOOST_AUTO_TEST_CASE(TEST, *boost::unit_test::timeout(240)) {
    BOOST_TEST_INFO(1);
}
```

DueTime is 64bit, but `static_cast<signed long int>(p_timeout) * 10` will 32bit and timeout value will overflow.


raffienficiaud:
Replaced with `LONGLONG` instead. Currently building in the branch `topic/PR-232-fix-timeout-in-windows`. Thanks!

raffienficiaud:
Merged to master, thanks!

PR #247, opened 2020-01-03
Use __linux__ instead of __linux.

giomasce:
The macro __linux is less portable and, for example, is not defined on
some architectures.

See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=28314.


raffienficiaud:
In master, thanks!

Issue #254, opened 2020-02-23
Console colors are not restored on Windows


valiko-ua:
Failing tests set color to red and do not restore (on Windows), so that all subsequent output is red.

Program to reproduce:

```c++
#define BOOST_TEST_MODULE My Test
#include <boost/test/included/unit_test.hpp>
BOOST_AUTO_TEST_CASE(my_test)
{
    int i = 1;
    BOOST_TEST(i == 2);
    BOOST_TEST(i == 3);
}
```
![boost_test_bug_win](https://user-images.githubusercontent.com/13869060/75120442-baf52d00-5694-11ea-8c6d-3684e1afcf34.png)

raffienficiaud:
Fix on branch `topic/GH-254-color-restored-windows`. Would be good if you give a try.

valiko-ua:
I tested on Windows: commit https://github.com/boostorg/test/commit/63ab1f2bd67bb80ba8512a560fd42065295151c0 fixes this bug. Colors are restored.

raffienficiaud:
Fixed in master

PR #255, opened 2020-02-23
Restore console colors after failing assertions on Windows

valiko-ua:
Fix for #254


valiko-ua:
With this fix the output is correct:

boost_test_win_fix

valiko-ua:
Unfortunately this fix works only for header-only mode.

PR #259, opened 2020-03-07
Avoid deprecated bind placeholders in global namespace

mloskot:
This fixes numerous yet annoying compilation warnings:

```
  note: #pragma message:
    The practice of declaring the Bind placeholders (_1, _2, ...)
    in the global namespace is deprecated.
    Please use <boost/bind/bind.hpp> + using namespace boost::placeholders,
    or define BOOST_BIND_GLOBAL_PLACEHOLDERS to retain the current behavior.
```

Apart from obscuring the output of compilation of libraries using Boost.Test, it may be troublesome on CI builds. On many CI services every line of log counts and if log dumps become too long, 
builds are terminated.

### DIFF
--- a/include/boost/test/data/test_case.hpp
+++ b/include/boost/test/data/test_case.hpp
@@ -32,7 +32,7 @@
 #include <boost/preprocessor/control/iif.hpp>
 #include <boost/preprocessor/comparison/equal.hpp>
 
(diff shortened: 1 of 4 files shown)


raffienficiaud:
In master

mloskot:
Thank you

PR #267, opened 2020-04-20
Use macOS API only on macOS

pinotree:
`__MACH__` matches any Mach-based OS, which means both macOS and GNU/Hurd; since the API used is really specific to macOS, then also check for `__APPLE__`.

This fixes the build on GNU/Hurd.


raffienficiaud:
in master

PR #270, opened 2020-05-06
C++20 compatibility for print_log_value<>

avikivity:
operator<<(std::ostream&, const wchar_t*) has been removed from C++20 [1].
Remove the corresponding print_log_value<> specialization.

### DIFF
--- a/include/boost/test/impl/test_tools.ipp
+++ b/include/boost/test/impl/test_tools.ipp
@@ -121,11 +121,15 @@ print_log_value<char const*>::operator()( std::ostream& ostr, char const* t )
 
 //____________________________________________________________________________//
 
+#if __cplusplus <= 201703L
+
 void
 print_log_value<wchar_t const*>::operator()( std::ostream& ostr, wchar_t const* t )
 {
     ostr << ( t ? reinterpret_cast<const void*>(t) : "null string" );
 }
+
+#endif
 
 //____________________________________________________________________________//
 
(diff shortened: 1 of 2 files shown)


avikivity:
I see some automated checks failed, but as far as I can tell these failures are not related to my changes.

raffienficiaud:
The question is rather to know if this change is relevant: `print_log_value` is the entity that prints any type to the logger. Removing this would mean that we are not supporting anymore logging `wchar_t const*`.

While the current implementation is incorrect (we just print the address of the string) it still does not prevent something like `BOOST_TEST(a == b)` to compile. 

This is a recurrent topic now, and it needs to be addressed. The "how" is unclear to me but my approach would not be the one you are proposing.

raffienficiaud:
> I see some automated checks failed, but as far as I can tell these failures are not related to my changes.

Yes this is also a recurrent topic :) I restarted some of those Appveyor is not yet working.

avikivity:
> The question is rather to know if this change is relevant: `print_log_value` is the entity that prints any type to the logger. Removing this would mean that we are not supporting anymore logging `wchar_t const*`.
> 
> While the current implementation is incorrect (we just print the address of the string) it still does not prevent something like `BOOST_TEST(a == b)` to compile.
> 
> This is a recurrent topic now, and it needs to be addressed. The "how" is unclear to me but my approach would not be the one you are proposing.

Actually the bug was already fixed in 8ac88c62dcc809d42daf8b6bef10f7adecc46dd1. I copied my hack without noticing that the code already changed. Sorry about the noise.

raffienficiaud:
Thanks for the quick update!

PR #278, opened 2020-07-10
Use boost_test_print_type customization point in lazy_ostream

ecatmur:
This makes BOOST_TEST_MESSAGE(x) consistent with BOOST_TEST(x).


raffienficiaud:
Merged through #335 rev c280b79

raffienficiaud:
Thanks for the patch!

PR #281, opened 2020-08-25
Export execution_aborted exception to allow catching it from outside the DLL

pkl97:
In our code we have encapsulated testcase execution into a dedicated method where we do some housekeeping.

The code looks like this:

```
    try
    {
        // execute the testcase
    }
    catch (const boost::execution_aborted&) {
        throw; // we let this exception propagate to ensure Boost.Test works as designed
    }
    catch (...) { // one of our exceptions
        // we do some housekeeping and then call BOOST_FAIL()
    }

```

However to let boost::execution_aborted exceptions propagate it is necessary for the type to be exported from the DLL. Otherwise the exception simply cannot be caught in the main program and always ends up in the catch-all handler.


raffienficiaud:
Merged through #330 with rev 8b54099e15d63e8fe4127c75c9643d5fd4a956d0

raffienficiaud:
Thanks for the patch!

PR #275, opened 2020-06-17
Replace anonymous enum arity with static constexpr

ecatmur:
https://github.com/boostorg/test/issues/268

PR #333, opened 2022-03-05
Topic/gh 275 anonymous enum arity replacement

raffienficiaud:
Build for #275


include/boost/test/data/monomorphic/initializer_list.hpp
```diff
@@ -75,7 +75,7 @@ class init_list<bool> {
 public:
     typedef bool sample;
 
-    enum { arity = 1 };
+    BOOST_STATIC_CONSTEXPR std::size_t arity = 1;
```
ecatmur, line 78:
```suggestion
    static const int arity = 1;
```

ecatmur:
Yes, this looks good. I won't be able to test it in our code base till Monday, but I can't see any reason it wouldn't work.

PR #351, opened 2022-08-30
Fix -Wformat on 32-bit Androids

georgthegreat:
This fixes a bunch of these errors:

```
In file included from boost/test/include/boost/test/included/unit_test.hpp:23:
boost/test/include/boost/test/impl/execution_monitor.ipp:429:27: error: format specifies type 'unsigned long' but the argument has type 'uintptr_t' (aka 'unsigned int') [-Werror,-Wformat]
                          (uintptr_t) m_sig_info->si_addr );
```


mborland:
Fixed by: https://github.com/boostorg/test/pull/353

PR #332, opened 2022-03-04
Add GitHub Actions config

raffienficiaud:
Isolated PR for checking #303

PR #361, opened 2022-10-10
Add Github Actions CI

Flamefire:
Add CI via Github actions and Boost.CI

Closes #332 
Closes #303

Currently it fails on Windows due to a timeout in test-timeout-suite.cpp (I guess running 6 tests with a sleep of 100ms already gets over 1s) and on OSX where the custom RTTI implementation fails.


mborland:
Superseded by #370

PR #367, opened 2022-12-19
fix: exception during startup

solosTec:
On my machine with windows 10 and the latest version of visual studio I get an exception during startup. The reason is that normalize_test_case_name() creates an ill-formed string object. The provided change in line 438 fixes the problem.


mborland:
#374 has been merged with the fix.

PR #375, opened 2023-02-22
Use BOOST_DEFAULTED_FUNCTION on empty destructors

ecatmur:
The compiler-generated copy constructor and copy assignment operator are deprecated since C++11 on classes with user-declared destructors.

This change allows clean compilation with the -Wdeprecated-copy-dtor/-Wdeprecated-copy-with-user-provided-dtor flag.


mborland:
This looks correct to me. I'll check again once the CI system cycles.

PR #382, opened 2023-04-25
Fix check_is_close_t for C++23 `<stdfloat>` types

mborland:
Implicit conversions are ill-formed so make them explicit


mborland:
Ubuntu 18.04 has been removed from Github actions so need to use containers. Since this is a typedef and a cast I think it is safe for now.

PR #390, opened 2023-07-18
Fix MSVC compiler warning

MarcelRaad:
Remove `std::move` on a temporary object to fix the following level 4 warning with Visual Studio 17.4 and later:

warning C5263: calling 'std::move' on a temporary object prevents copy elision


mborland (approved):
LGTM. Thanks

PR #393, opened 2023-08-14
fix unreachable return compilation error

timmaraju:
On Visual Studio 16.9, the return was marked as unreachable because of the throw before


mborland (approved):
LGTM. I approved the CI run so we'll see what that has to say.

timmaraju:
@mborland Sorry for coming back so late to this, can we start the CI run again on this please?

mborland:
> @mborland Sorry for coming back so late to this, can we start the CI run again on this please?

The only issue was with 4.6 which has been dropped since this PR opened.

Issue #398, opened 2023-11-13
Overflow warning in `print_helper.hpp`


JBouwer:
Using:
``
clang++   -stdlib=libc++ -std=c++17 -Wall ...
``
on an Intel x64 Mac (Clang 11), with selected code [^1]
results in:
```
...
In file included from boost/test/unit_test.hpp:18:
In file included from boost/test/test_tools.hpp:42:
In file included from boost/test/tools/context.hpp:19:
In file included from boost/test/utils/lazy_ostream.hpp:16:
boost/test/tools/detail/print_helper.hpp:100:71: warning: overflow in expression; result is 2147483347 with type 'int' [-Winteger-overflow]
            return ostr.precision( 2 + std::numeric_limits<T>::digits * 301/1000 );
                                                                      ^
```

It is pretty easily fixed by promoting the calculation to the parameter type;
``
            return ostr.precision( 2 + std::streamsize(std::numeric_limits<T>::digits) * 301/1000 );
``

 [^1]: Occurences in my code are pretty obscure:
For me it manifests during (implicit) printing of big enough types;
I can induce it with:
``
    BOOST_CHECK_NE( boost::multiprecision::checked_cpp_int(std::numeric_limits<int>::max()), 0 );
``

PR #399, opened 2023-11-13
Fix overflow warning in `print_helper.hpp`. #398

JBouwer:
- (Implicitly) printing a big enough type results in: "warning: overflow in expression; result is 2147483347 with type 'int' [-Winteger-overflow]".
- Fixed by promoting the calculation to 'std::streamsize'.


mborland (approved):
LGTM. I approved the CI run as well.

PR #400, opened 2023-11-14
Remove C++03 Testing

mborland:


raffienficiaud:
Hi Matt, if I am not mistaken there is a announcement to make on Boost mailing list so that Boost.Test can drop support of C++03. Maybe this has been done already, I have not checked.

mborland:
> Hi Matt, if I am not mistaken there is a announcement to make on Boost mailing list so that Boost.Test can drop support of C++03. Maybe this has been done already, I have not checked.

You are correct. I noticed on this PR's CI run that a number of the upstream libraries removed C++03 support: https://github.com/boostorg/test/pull/399. I'll add a warning in the release that is happening now that in the next release C++03 will be gone.

PR #403, opened 2023-11-29
fix "boost/test/impl/execution_monitor.ipp" compilation in GCC 13.2

gh-user-2022:
Without this fix GCC was giving compile error that PRIxPTR is not defined.

### DIFF
--- a/include/boost/test/impl/execution_monitor.ipp
+++ b/include/boost/test/impl/execution_monitor.ipp
@@ -202,7 +202,7 @@ namespace { void _set_se_translator( void* ) {} }
 #      define BOOST_TEST_DEFINED_STDC_FORMAT_MACROS
 #    endif
 #  endif
-#  include <inttypes.h>
+#  include <cinttypes>
 #  define BOOST_TEST_PRIxPTR PRIxPTR
 #  ifdef BOOST_TEST_DEFINED_STDC_FORMAT_MACROS
 #    undef __STDC_FORMAT_MACROS


mborland:
Do you have a reproducer? I just ran the tests on Fedora 39 with gcc version 13.2.1 20231011 (Red Hat 13.2.1-4) with no issues. Godbolt shows the macro is defined: https://godbolt.org/z/3ETsozcz6

gh-user-2022:
In CentOS 6 when I build the boost-1.83 with GCC 13.2 there is error:

```
gcc.compile.c++ bin.v2/libs/test/build/gcc-13/release/cxxstd-20-iso/link-static/threading-multi/visibility-hidden/execution_monitor.o
In file included from libs/test/src/execution_monitor.cpp:16:
./boost/test/impl/execution_monitor.ipp: In member function `void boost::detail::system_signal_exception::report() const':
./boost/test/impl/execution_monitor.ipp:435:88: error: expected `)' before `PRIxPTR'
  435 |                           "signal: illegal opcode; address of failing instruction: 0x%" BOOST_TEST_PRIxPTR,
      |                                                                                        ^
      |                                                                                        )
./boost/test/impl/execution_monitor.ipp:434:25: note: to match this `('
  434 |             report_error( execution_exception::system_fatal_error,
      |                         ^
./boost/test/impl/execution_monitor.ipp:205:1: note: `PRIxPTR' is defined in header `<cinttypes>'; did you forget to `#include <cinttypes>'?
  204 | #include <boost/test/detail/suppress_warnings.hpp>
  +++ |+#include <cinttypes>
```

mborland:
That diff will be in Boost 1.84 which releases in 2 weeks.

PR #406, opened 2023-12-16
Respect library-specific _NO_LIB macros. Fixes #339.

### LINKED ISSUE #339: Library-specific _NO_LIB macros should be recognized and disable autolink
The library-specific _DYN_LINK macros defined by the CMake config files are recognized (https://github.com/boostorg/test/commit/625bafd2cdc3e9a6f338573a17558050e2bd0e58) but the library-specific _NO_LIB macros are not.

This used to not matter because the config files used to define BOOST_ALL_NO_LIB, but they define the library-specific _NO_LIB macros now.

Ref: https://lists.boost.org/Archives/boost/2022/06/253153.php

pdimov:


raffienficiaud:
LGTM, @pdimov do you confirm this fixes your issue #339 ?

pdimov:
Yes, the PR is mine, I tested it locally, and it seems to work.

PR #411, opened 2024-02-29
Fix boost::function streaming error

mborland:
@raffienficiaud The PR for your patch


raffienficiaud:
Thank you @mborland . I just pushed a tiny warning removal.

PR #426, opened 2024-07-20
Add support for modular build structure.

grafikrobot:
This is part of the effort to make the Boost libraries "modular" for build and consumption. See https://lists.boost.org/Archives/boost/2024/01/255704.php and https://github.com/grafikrobot/boost-b2-modular/blob/b2-modular/README.adoc for more information.

This PR depends on the following other PRs being merged to both develop and master branches of the respective repos:

- https://github.com/boostorg/boost/pull/854

This PR will be changed to ready for review, i.e. not draft, when the above are merged. Do not merge this one until that time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,13 +110,13 @@ jobs:
             install:
               - g++-12-multilib
           - name: UBSAN
-            toolset: gcc-12
-            cxxstd: "17"
+            toolset: gcc-13
+            cxxstd: "11"
             address_model: 64
             ubsan: 1
-            os: ubuntu-22.04
+            os: ubuntu-24.04
             install:
-              - g++-12-multilib
+              - g++-13-multilib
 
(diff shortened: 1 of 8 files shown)


raffienficiaud:
I am seeing a "no space left on device" error.

mborland:
> > @grafikrobot Do the modular changes come with an increase in memory usage? The only failure in the CI is OOM with UBSAN that I have not seen before.
> 
> They don't as they are build system changes only. But if those tests where close to OOM before it's not surprising that it might OOM at any time.

Build system changes could definitely change the memory consumption. Since test has quite a few dependencies are the build sizes all the same? Were build sizes tested before and after modularity to ensure size parity? Is each dependency dragging in copies of all it's dependencies, or is it like how it used to be where there's just one copy of everything?

fanquake:
@grafikrobot Is there any further update on your investigations here?

PR #432, opened 2024-10-15
Rename method arguments to avoid collision with linux UAPI macro

notdanhan:
`tc_gen` is a macro in the linux UAPI, if included before `<boost/test/unit_test.hpp>` it causes compilation to fail.


mborland (approved):
Looks reasonable; I approved the CI run.

Issue #436, opened 2025-01-17
Build error: `execution_monitor.ipp: error: expected ')' before 'PRIxPTR'`


barracuda156:
The following header has an inaccurate condition to define `__STDC_FORMAT_MACROS`:

https://github.com/boostorg/test/blob/fcd634a534835068c2f160101cd19c23b7360dcb/include/boost/test/impl/execution_monitor.ipp#L197-L212

So I get this failure on macOS, for example:
```
:info:build ./boost/test/impl/execution_monitor.ipp:444:25: note: to match this '('
:info:build   444 |             report_error( execution_exception::system_fatal_error,
:info:build       |                         ^
:info:build ./boost/test/impl/execution_monitor.ipp:215:1: note: 'PRIxPTR' is defined in header '<cinttypes>'; this is probably fixable by adding '#include <cinttypes>'
:info:build   214 | #include <boost/test/detail/suppress_warnings.hpp>
:info:build   +++ |+#include <cinttypes>
:info:build   215 | 
:info:build ./boost/test/impl/execution_monitor.ipp:445:86: warning: spurious trailing '%' in format [-Wformat=]
:info:build   445 |                           "signal: illegal opcode; address of failing instruction: 0x%" BOOST_TEST_PRIxPTR,
:info:build       |                                                                                      ^
:info:build ./boost/test/impl/execution_monitor.ipp:445:27: warning: too many arguments for format [-Wformat-extra-args]
:info:build   445 |                           "signal: illegal opcode; address of failing instruction: 0x%" BOOST_TEST_PRIxPTR,
:info:build       |                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
:info:build ./boost/test/impl/execution_monitor.ipp:450:86: error: expected ')' before 'PRIxPTR'
:info:build   450 |                           "signal: illegal trap; address of failing instruction: 0x%" BOOST_TEST_PRIxPTR,
:info:build       |                                                                                      ^
:info:build       |                                                                                      )
:info:build ./boost/test/impl/execution_monitor.ipp:449:25: note: to match this '('
:info:build   449 |             report_error( execution_exception::system_fatal_error,
:info:build       |                         ^
:info:build ./boost/test/impl/execution_monitor.ipp:206:30: note: 'PRIxPTR' is defined in header '<cinttypes>'; this is probably fixable by adding '#include <cinttypes>'
[...]

PR #437, opened 2025-01-22
Add additional paths to guarantee BOOST_TEST_PRIxPTR is defined

mborland:
Closes: #436


barracuda156:
@mborland I cannot re-open the issue, but this patch does not fix the problem, since the new fallback, while desirable, is not being used.

PR #438, opened 2025-01-26
execution_monitor.ipp: follow-up fix for macOS

barracuda156:
@mborland This fixes https://github.com/boostorg/test/issues/436


mborland (approved):
Thanks for fixing my fix

PR #441, opened 2025-02-03
Add test for the size_t class

Flamefire:
Verify the constructor and increment/decrement operators

I wasn't sure https://github.com/boostorg/test/blob/10185327515e968ab01c9a5a4017c603777e4cd7/include/boost/test/data/size.hpp#L51 works as expected as the templated ctor might be preferred which would not copy the `is_inf`

The test also shows some inconsistencies in the over/underflow handling. It might be better to remove any operator taking a `std::size_t` which means those values will get converted to `data::size_t` and can be consistently handled.

### DIFF
--- a/include/boost/test/data/size.hpp
+++ b/include/boost/test/data/size.hpp
@@ -30,7 +30,7 @@ namespace data {
 // **************                    size_t                    ************** //
 // ************************************************************************** //
 
-//! Utility for handling the size of a datasets
+//! Utility for handling the size of a dataset
 class size_t {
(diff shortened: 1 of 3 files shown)


mborland (approved):
LGTM

mborland:
> The test also shows some inconsistencies in the over/underflow handling. It might be better to remove any operator taking a `std::size_t` which means those values will get converted to `data::size_t` and can be consistently handled.

Yeah, that would likely be an improvement.

PR #480, opened 2026-02-26
Ignore c2y-extensions warning with Clang >= 22

### LINKED ISSUE #479: clang 22.1.0 dislikes the use of __COUNTER__ with cpp standard 23
When compiling tests with boost test and clang 22.1.0 from the llvm apt repository, clang emits a lot of `Wc2y-extensions` warnings around boost::test's use of the `__COUNTER__` macro.

While I've added `-Wno-c2y-extensions` to my project to silence this, can the fallback of __LINE__ be used if the c++ standard is set at 23?

mborland:
Closes: #479


Flamefire:
Looks like this isn't enough: The warning needs to be suppressed where the macro is expanded. E.g. https://fuchsia.googlesource.com/third_party/benchmark/+/d8db2f90b643eb28a12976beb4d57bcfb639911d%5E%21/#F2

